PR #3, opened 2024-04-12
Add New Persistent Response System + Remove A Couple Dependencies

obelisk:
## Persistent Response
To fix some performance issues with handling GET requests via rule invocation, this adds a new system: persistent response. This is not the same persistence as the database system, but persistent in that rules can now set a response, and that response will still be available the next time they are invoked. There is a new API call in the STL to fetch this data called `get_response`.

This system primarily exists so that you can configure GET handlers to return this persistent response instead of resulting in a spawn blocking call into the Wasmer runtime. The idea is that cache invalidation is now handled by the asynchronous nature of further messages coming in and the rule modifying the response.

As an example, let's say you have a rule processing a bunch of logs and you want a way of asking the rule what the largest log it's seen is? Before your rule would need to store this in the persistent database, then handle the call into the rule for the LogSource::Get to just return the largest log.

Now you can continue processing logs as normal and store the largest log in the DB, but also return it as a response. When a get request comes in, it will immediately return the last response, rather than calling your rule.

### Cold Start 
When Plaid starts however, this field will not be populated. It might not even be populated after processing some logs depending on how your rule/module works. To address this, you can configure the caching mode with "call_on_none". When set to true, if a GET request comes in and there is no persistent log to return, your module *will* be invoked (with the standard LogSource::Get) to give it a chance to run. In this case, the response set by the rule will be saved both to the persistent response AND returned to service that get request.

## Dependencies

In light of the `xz` exploit I'm going through and trying to remove any dependencies we might not need. This has the added benefit of cleaning and speeding up the build.

Removing futures was pretty easy but will need some soak testing since it's concurrency related (I think this is equivalent but need to make sure).

`Rand` was introduced by the new interval data generator but just to calculate splays. Since we have `SystemRandom` from `ring`, I reimplemented a simple, not cryptographically secure, random range.


plaid/src/main.rs
```diff
@@ -83,15 +82,15 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    let webhook_servers: Vec<Box<Pin<Box<_>>>> = config
         .webhooks
         .into_iter()
         .map(|(server_name, config)| {
             let log_sender = log_sender.clone();
             let server_address: SocketAddr = config
                 .listen_address
                 .parse()
-                .expect("{server_name} had an invalid address");
+                .expect("A server had an invalid address");
```
timweri, line 93: Just curious. Is there a reason for this?

obelisk: Server name never worked

plaid/Cargo.toml
```diff
@@ -26,7 +23,7 @@ paste = "1.0"
 plaid_stl = { path = "../plaid-stl" }
 rcgen = { version = "0.10", features = ["x509-parser"] }
 regex = "1"
-ring = "0.16.20"
+ring = "0.17"
```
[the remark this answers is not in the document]
obelisk: I like 0.17 because it allows patches to come in

plaid/src/loader/mod.rs
```diff
@@ -140,6 +192,14 @@ pub fn load(config: Configuration) -> Result<PlaidModules, ()> {
             type_[0].to_string()
         };
 
+        // See if a type is defined in the configuration file, if not then we will grab the first part
+        // of the filename up to the first underscore.
```
timweri, line 196: nit: This needs to be rewritten

plaid/src/main.rs
```diff
@@ -168,81 +170,110 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                                    accessory_data: query.into_iter().map(|(k, v)| (k, v.into_bytes())).collect(),
+                                    source,
+                                    logbacks_allowed,
+                                };
+                                // We need to spawn blocking here because the executor is not async
+                                // and it will need to block inside the host API functions to call other async
+                                // functions on a different runtime.
+                                //
+                                // As far as I can tell there is no way around this since WASM host functions
+                                // cannot be async. I am afraid this prevents a denial of service risk because of
```
timweri, line 243:
```suggestion
                                // cannot be async. I am afraid this presents a denial of service risk because of
```

plaid/src/loader/mod.rs
```diff
@@ -35,16 +35,68 @@ pub struct Configuration {
...
+    pub fn get_data(&self) -> String {
+        match self.data.read() {
+            Ok(data) => data.clone(),
+            Err(e) => {
+                error!(
+                    "Critical error getting a read lock on persistent response: {:?}",
+                    e
+                );
+                String::new()
+            }
```
timweri, line 70: Could you explain the design choice here to not return `Result<String>`?

PR #5, opened 2024-07-09
Okta tailer 

wbssbw:
This PR fixes the current issues with the Okta data generator and implements additional customizability to the generator.

The generator was failing due to an issue in the URI string where the date provided did not match the requirements set by the Okta `/log` API. To address this, we convert the timestamp to RFC3339 before sending the request.

We also add additional configuration values for the generator in the form of `limit` and `sleep_duration`. `limit` is passed as a query param to the API to limit the number of logs returned from Okta. If no value is provided, we use Okta's default of 100. `sleep_duration` is intended to mitigate any possibility of getting rate limited by Okta. The `/logs` endpoint restricts us to 50 calls/minute and consequently, it's rate limiting is likely. After successfully fetching and forwarding logs to the execution system, we sleep for this amount of time in between calls. If no value is provided, we will wait for 1 millisecond

**Other changes:**
I also added an Optional configurable timeout duration for the General, Okta, PagerDuty, Slack, Splunk, and Yubikey API implementations. If no timeout duration is provided, Plaid just defaults to 5 seconds. I've been running into some instances lately where requests time out and this is an easy way to give more control to the Plaid admin on how their system should run

### DIFF
--- a/plaid/src/apis/general/mod.rs
+++ b/plaid/src/apis/general/mod.rs
@@ -12,14 +12,15 @@ use std::time::Duration;
 
 use crate::{data::DelayedMessage, executor::Message};
 
-use super::DEFAULT_TIMEOUT_SECONDS;
+use super::default_timeout_seconds;
 
 #[derive(Deserialize)]
 pub struct GeneralConfig {
     network: network::Config,
     /// The number of seconds until an external API request times out.
-    /// If `None`, the `DEFAULT_TIMEOUT_SECONDS` will be used.
-    api_timeout_seconds: Option<u64>,
+    /// If no value is provided, the result of `default_timeout_seconds()` will be used.
+    #[serde(default = "default_timeout_seconds")]
+    api_timeout_seconds: u64,
 }
 
 pub struct General {
@@ -36,11 +37,8 @@ impl General {
         log_sender: Sender<Message>,
         delayed_log_sender: Sender<DelayedMessage>,
     ) -> Self {
-        let timeout_seconds = config
-            .api_timeout_seconds
-            .unwrap_or(DEFAULT_TIMEOUT_SECONDS);
         let client = reqwest::Client::builder()
-            .timeout(Duration::from_secs(timeout_seconds))
+            .timeout(Duration::from_secs(config.api_timeout_seconds))
             .build()
             .unwrap();
 
--- a/plaid/src/apis/general/mod.rs
+++ b/plaid/src/apis/general/mod.rs
@@ -12,9 +12,14 @@ use std::time::Duration;
 
 use crate::{data::DelayedMessage, executor::Message};
 
+use super::DEFAULT_TIMEOUT_SECONDS;
+
 #[derive(Deserialize)]
 pub struct GeneralConfig {
     network: network::Config,
+    /// The number of seconds until an external API request times out.
+    /// If `None`, the `DEFAULT_TIMEOUT_SECONDS` will be used.
+    api_timeout_seconds: Option<u64>,
 }
 
 pub struct General {
@@ -31,8 +36,11 @@ impl General {
         log_sender: Sender<Message>,
         delayed_log_sender: Sender<DelayedMessage>,
     ) -> Self {
+        let timeout_seconds = config
+            .api_timeout_seconds
+            .unwrap_or(DEFAULT_TIMEOUT_SECONDS);
         let client = reqwest::Client::builder()
-            .timeout(Duration::from_secs(5))
+            .timeout(Duration::from_secs(timeout_seconds))
             .build()
             .unwrap();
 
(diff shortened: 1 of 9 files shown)


plaid/src/apis/general/mod.rs
```diff
@@ -12,9 +12,14 @@ use std::time::Duration;
...
 use crate::{data::DelayedMessage, executor::Message};
 
+use super::DEFAULT_TIMEOUT_SECONDS;
+
 #[derive(Deserialize)]
 pub struct GeneralConfig {
     network: network::Config,
+    /// The number of seconds until an external API request times out.
+    /// If `None`, the `DEFAULT_TIMEOUT_SECONDS` will be used.
+    api_timeout_seconds: Option<u64>,
```
obelisk, line 22: I wouldn't make this an option, use the serde deserialize with default feature. Then you won't have to unwrap this eveyrwhere.

plaid/src/data/okta/mod.rs
```diff
@@ -1,59 +1,107 @@
...
-use time::{format_description::well_known::Rfc3339, OffsetDateTime};
-
 use std::time::Duration;
-
-use crate::executor::Message;
+use time::{format_description::well_known::Rfc3339, OffsetDateTime};
 
 #[derive(Deserialize)]
 pub struct OktaConfig {
+    /// API token used to authenticate to Okta
```
obelisk, line 12: Thank you

plaid/src/data/okta/mod.rs
```diff
@@ -1,59 +1,107 @@
...
+    /// Sending channel used to send logs into the execution system
     logger: Sender<Message>,
 }
 
 /// We try not to parse anything complicated since our job is just
 /// to pass it on.
+/// See https://developer.okta.com/docs/reference/api/system-log/#logevent-object for full docs
 #[derive(Deserialize, Serialize)]
 struct OktaLog {
+    /// Timestamp when the event is published
```
obelisk, line 66: The hero we need but not the one we deserve

plaid/src/data/okta/mod.rs
```diff
@@ -1,59 +1,107 @@
...
     legacy_event_type: Value,
+    /// The transaction details of an action
     transaction: Value,
+    /// Unique identifier for an individual event
     uuid: Value,
+    /// Versioning indicator
     version: Value,
+    /// The request that initiates an action
     request: Value,
+    /// Zero or more targets of an action
```
obelisk, line 104: Does this struct generate a whole bunch of unused warnings? If we don't need to parse it maybe we should just hide it behind a `Value` and pull out the published field manually

obelisk (changes requested):

plaid/src/data/okta/mod.rs
```diff
@@ -1,89 +1,111 @@
...
+    /// Filters the lower time bound of the log events published property for bounded queries or persistence time for polling queries
     since: OffsetDateTime,
+    /// Sending channel used to send logs into the execution system
     logger: Sender<Message>,
 }
 
-/// We try not to parse anything complicated since our job is just
-/// to pass it on.
-#[derive(Deserialize, Serialize)]
-struct OktaLog {
```
obelisk, line 34: Red diffs are the best diffs

plaid/src/data/okta/mod.rs
```diff
@@ -92,63 +114,103 @@ impl Okta {
...
-                        ))
-                        .unwrap();
-                }
+                // Send log into logging system to be processed by rule(s)
+                //
+                // Eventually these errors need to bubble up so the service can shut down
+                // then be restarted by an orchestration service
+                self.logger
+                    .send(Message::new(
+                        "okta".to_string(),
```
obelisk, line 199: Nice

plaid/src/data/okta/mod.rs
```diff
@@ -92,63 +114,103 @@ impl Okta {
...
+                if log_timestamp > most_recent_log_seen {
+                    most_recent_log_seen = log_timestamp;
                 }
 
-                if log_timestamp < self.since {
-                    self.since = newest_timestamp.unwrap();
-                    return Ok(());
-                }
+                // Attempts to parse the log received from Okta to bytes.
+                let log_bytes = match serde_json::to_vec(&log) {
```
obelisk, line 185: I think this is wrong. The type we have now is a single log. Not a vec.

plaid/src/data/okta/mod.rs
```diff
@@ -92,63 +114,103 @@ impl Okta {
...
-                        .unwrap();
-                }
+                // Send log into logging system to be processed by rule(s)
+                //
+                // Eventually these errors need to bubble up so the service can shut down
+                // then be restarted by an orchestration service
+                self.logger
+                    .send(Message::new(
+                        "okta".to_string(),
+                        log_bytes,
```
obelisk, line 200: I think this should be serializing out the Value to a string then getting the bytes of that

plaid/src/data/okta/mod.rs
obelisk: I think this should be to_string

obelisk: Nope, @wbssbw is right as far as I can tell

obelisk (approved):

PR #11, opened 2024-08-07
Config verification

wbssbw:
Adds a simple binary that interpolates secrets into the config file and prints it, along with its hash to stdout. This is useful for verifying that a config with secrets interpolated is the same as a config without one - allowing us to say with 100% certainty that configs are correct.

The hashing is a nice to have IMO. I can see wanting to remove it to cut down on dependencies (I chose to use sha3 because it has the fewest dependencies). We can easily just copy and paste the config printed to stdout into any online hasher and achieve the same result.


plaid/src/bin/config_check.rs
```diff
@@ -0,0 +1,52 @@
...
+        .expect("Failed to read secrets file");
+
+    let secret_map = serde_json::from_slice::<Value>(&secrets)
+        .unwrap()
+        .as_object()
+        .cloned()
+        .unwrap();
+
+    // Iterate over the secrets we just parsed and replace matching keys in the config
+    for (secret, value) in secret_map {
```
obelisk, line 38: Can we abstract this into a function call with shared code in the Plaid library itself? That was as we update that code, it will apply to config check as well>

obelisk (changes requested):
This should also verify the output config is valid for Plaid

plaid/Cargo.toml
```diff
@@ -38,6 +38,7 @@ warp = { version = "0.3", features = ["tls"] }
 wasmer = { version = "4", default-features = false, features = ["cranelift"] }
 wasmer-middlewares = "4"
 jsonwebtoken = { version = "9.2" }
+sha3 = "0.10.8"
```
obelisk, line 41: Why import SHA3 for this? You could do a SHA2 from Ring without adding a new dependency. Is there a reason you want to pull in a new dep for this?

plaid/src/bin/config_check.rs
```diff
@@ -0,0 +1,27 @@
+use clap::{Arg, Command};
+use plaid::config::read_and_interpolate;
+
+fn main() {
+    env_logger::init();
+    let matches = Command::new("Config Check")
```
obelisk, line 6: I tried running help for this but couldn't get it to function. Not sure if there is something wrong with the invocation but would be nice to get help to show up by default.

obelisk (changes requested):

obelisk (approved):

PR #16, opened 2024-08-13
Implement WebSocket data generator

wbssbw:
This PR implements WebSocket data generators to Plaid. 

The `websocket` module defines an optional `WebsocketDataGenerator` configuration struct that takes map of generator names to socket configuration. This allows us to give names to the generators that will later by passed to rules.

In the example configuration below, we define a generator named `demo_rpc_call`. Its fields are:
- `uris`: A list of URIs that can generate data. We allow for many URIs to add resiliency in the case that a socket becomes disconnected. The generator implements a round-robin strategy with exponential backoff to retry connections to sockets on failure.
    - There are two additional fields for further customization of the retry strategy: (1) `min_retry_duration` and (2) `max_retry_duration`. As their names suggest, these are the minimum and maximum amount of time to wait before retrying connections. If no values are provided for either `min_retry_duration` defaults to 100 milliseconds and `max_retry_duration` defaults to 60000 milliseconds (1 minute).
- `message`: This is the data sent in the message to the socket. This field is `Option` in the event that you only need to read from a socket.
- `sleep_duration`: The time to wait in between sending new messages t the socket
- `headers`: Optional headers to include in the request. _Note: These headers will NOT overwrite headers required for the socket connection (e.g. `sec-websocket-key`)_.

```toml
[data.websocket]
[data.websocket.websockets]
[data.websocket."websockets"."demo_rpc_call"]
uris = ["wss://some_rpc"]
message = "{ \"id\": 1, \"jsonrpc\": \"2.0\", \"method\": \"eth_getBlockByNumber\", \"params\": [ \"finalized\", false ]}"
log_type = "testing"
sleep_duration = 0
[data.websocket."websockets"."demo_rpc_call"."headers"]
"some-custom-header" = "asdf"
"another-custom-header" = "asdf"
```

**Tasks**
Each WebSocket spawns up to two tasks: a read task and a write task. If no message is provided in the configuration, the write task is not spawned. If either of these tasks returns, the connection to the socket has closed and we must restart.

### DIFF
--- a/plaid-stl/src/messages.rs
+++ b/plaid-stl/src/messages.rs
@@ -5,7 +5,7 @@ pub enum Generator {
     Github,
     Okta,
     Interval(String),
-    Websocket(String),
+    WebSocketExternal(String),
 }
 
 impl std::fmt::Display for Generator {
@@ -14,7 +14,7 @@ impl std::fmt::Display for Generator {
             Generator::Github => write!(f, "github"),
             Generator::Okta => write!(f, "okta"),
             Generator::Interval(job) => write!(f, "interval/{job}"),
-            Generator::Websocket(ws) => write!(f, "websocket/{ws}"),
+            Generator::WebSocketExternal(ws) => write!(f, "websocket/{ws}"),
         }
     }
 }
--- a/plaid-stl/src/messages.rs
+++ b/plaid-stl/src/messages.rs
@@ -5,6 +5,7 @@ pub enum Generator {
     Github,
     Okta,
     Interval(String),
+    Websocket(String),
 }
 
 impl std::fmt::Display for Generator {
@@ -13,6 +14,7 @@ impl std::fmt::Display for Generator {
             Generator::Github => write!(f, "github"),
             Generator::Okta => write!(f, "okta"),
             Generator::Interval(job) => write!(f, "interval/{job}"),
+            Generator::Websocket(ws) => write!(f, "websocket/{ws}"),
         }
     }
 }
--- a/plaid/Cargo.toml
+++ b/plaid/Cargo.toml
@@ -38,6 +38,8 @@ warp = { version = "0.3", features = ["tls"] }
 wasmer = { version = "4", default-features = false, features = ["cranelift"] }
 wasmer-middlewares = "4"
 jsonwebtoken = { version = "9.2" }
+tokio-tungstenite = { version = "0.23.1", features = ["native-tls"] }
+futures-util = "0.3.30"
 
 # Uncomment to build with Quorum. This is needed
 # because otherwise cargo will try and find this
--- a/plaid/resources/plaid.toml
+++ b/plaid/resources/plaid.toml
@@ -119,6 +119,17 @@ secret_key = ""
 
 [data]
 
+# [data.websocket]
+# [data.websocket.websockets]
+# [data.websocket."websockets"."demo_rpc_call"]
+# log_type = "testing"
+# [data.websocket."websockets"."demo_rpc_call".uris]
+# simplystaking = "wss://some_websocket"
+# [data.websocket."websockets"."demo_rpc_call".message_config]
+# message = "{ \"id\": 1, \"jsonrpc\": \"2.0\", \"method\": \"eth_getBlockByNumber\", \"params\": [ \"finalized\", false ]}"
+# sleep_duration = 100
+# [data.websocket."websockets"."demo_rpc_call"."headers"]
+
 [webhooks."internal"]
 listen_address = "0.0.0.0:4554"
 [webhooks."internal".webhooks."AAAA"]
--- a/plaid/resources/plaid.toml
+++ b/plaid/resources/plaid.toml
@@ -127,7 +127,7 @@ secret_key = ""
 # simplystaking = "wss://some_websocket"
 # [data.websocket."websockets"."demo_rpc_call".message_config]
 # message = "{ \"id\": 1, \"jsonrpc\": \"2.0\", \"method\": \"eth_getBlockByNumber\", \"params\": [ \"finalized\", false ]}"
-# sleep_duration = 100
+# sleep_duration = 100 ## This means the message is sent every 0.1 seconds
 # [data.websocket."websockets"."demo_rpc_call"."headers"]
 
 [webhooks."internal"]
--- a/plaid/src/data/mod.rs
+++ b/plaid/src/data/mod.rs
@@ -167,7 +167,7 @@ impl Data {
             });
         }
 
-        info!("Started Data collection tasks");
+        info!("Started Data Generators");
         Ok(internal_sender)
     }
 }
--- a/plaid/src/data/mod.rs
+++ b/plaid/src/data/mod.rs
@@ -6,6 +6,7 @@ mod websocket;
 
 use crate::{
     executor::Message,
+    logging::Logger,
     storage::{Storage, StorageError},
 };
 
@@ -53,6 +54,7 @@ impl DataInternal {
         config: DataConfig,
         logger: Sender<Message>,
         storage: Option<Arc<Storage>>,
+        els: Logger,
     ) -> Result<Self, DataError> {
         let github = config
             .github
@@ -82,7 +84,7 @@ impl DataInternal {
 
         let websocket = config
             .websocket
-            .map(|ws| websocket::WebsocketGenerator::new(ws, logger.clone()));
+            .map(|ws| websocket::WebsocketGenerator::new(ws, logger.clone(), els));
 
         Ok(Self {
             github,
@@ -99,8 +101,9 @@ impl Data {
         config: DataConfig,
         sender: Sender<Message>,
         storage: Option<Arc<Storage>>,
+        els: Logger,
     ) -> Result<Option<Sender<DelayedMessage>>, DataError> {
-        let di = DataInternal::new(config, sender, storage).await?;
+        let di = DataInternal::new(config, sender, storage, els).await?;
         let handle = tokio::runtime::Handle::current();
 
         // Start the Github Audit task if there is one
--- a/plaid/src/data/mod.rs
+++ b/plaid/src/data/mod.rs
@@ -2,6 +2,8 @@ pub mod github;
 mod internal;
 mod interval;
 mod okta;
+mod websocket;
+
 use crate::{
     executor::Message,
     storage::{Storage, StorageError},
@@ -23,6 +25,7 @@ pub struct DataConfig {
     okta: Option<okta::OktaConfig>,
     internal: Option<internal::InternalConfig>,
     interval: Option<interval::IntervalConfig>,
+    websocket: Option<websocket::WebsocketDataGenerator>,
 }
 
 struct DataInternal {
@@ -34,6 +37,8 @@ struct DataInternal {
     internal: Option<internal::Internal>,
     /// Interval manages tracking and execution of jobs that are executed on a defined interval
     interval: Option<interval::Interval>,
+    /// Websocket manages the creation and maintenance of WebSockets that provide data to the executor
+    websocket: Option<websocket::WebsocketGenerator>,
 }
 
 pub struct Data {}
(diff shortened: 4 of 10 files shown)


plaid-stl/src/messages.rs
```diff
@@ -5,6 +5,7 @@ pub enum Generator {
     Github,
     Okta,
     Interval(String),
+    Websocket(String),
```
obelisk, line 8: Probably `WebSocket`

obelisk: We also might want to call this `WebSocketExternal` because in the future we may want to add the ability to host websockets for people to send logs to.

plaid/src/data/mod.rs
```diff
@@ -151,6 +160,13 @@ impl Data {
...
             });
         }
+
+        if let Some(websocket) = di.websocket {
+            handle.spawn(async move {
+                websocket.start().await;
+            });
+        }
+
         info!("Started Data collection tasks");
```
obelisk, line 170: Let's update this to `info!("Started Data Generators");` for consistency

plaid/src/data/websocket/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+use std::{collections::HashMap, str::FromStr, time::Duration};
+use tokio::{net::TcpStream, task::JoinHandle};
+use tokio_tungstenite::{
+    connect_async,
+    tungstenite::{client::IntoClientRequest, protocol::Message as WSMessage},
+    MaybeTlsStream, WebSocketStream,
+};
+
+#[derive(Debug)]
+enum Errors {
```
obelisk, line 22: Doc comments

plaid/src/data/websocket/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+    SocketCreationFailure,
+}
+
+#[derive(Deserialize)]
+pub struct WebsocketDataGenerator {
```
obelisk, line 27: Doc comments

plaid/src/data/websocket/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+    websockets: HashMap<String, Websocket>,
+}
+
+#[derive(Deserialize)]
+pub struct Websocket {
```
obelisk, line 32: Comment for struct

plaid/src/data/websocket/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+    /// The URI(s) of the WebSocket endpoint. The config allows multiple URIs in the event that one fails,
+    /// a new one can be used in its place.
```
obelisk, line 34: You will want to describe here how that other one is chosen. You may also want to define that as another option here. Possibilities, like `RoundRobin`, or `Random`, or some other metrics

plaid/src/data/websocket/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+    #[serde(deserialize_with = "parse_uris")]
+    uris: Vec<Uri>,
+    /// A string indicating the type of log associated with the WebSocket.
+    log_type: String,
+    /// The message to be sent over the WebSocket connection.
```
obelisk, line 39: I think this needs more detail. Why are we sending messages over the connection?

obelisk: How many times, in what circumstance?

plaid/src/data/websocket/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+}
+
+/// Custom parser for URI. Returns an error if an invalid URI is provided
+fn parse_uris<'de, D>(deserializer: D) -> Result<Vec<Uri>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let uris_raw = <Vec<String>>::deserialize(deserializer)?;
+
+    let uris = uris_raw
```
obelisk, line 74: Should we also check that the protocol is ws or wss?

plaid/src/data/websocket/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+
+        Self { clients }
+    }
+
+    /// Starts all WebSocket clients managed by this generator.
+    ///
+    /// This function initializes the WebSocket clients, logs the number of clients being initialized,
+    /// and then spawns an asynchronous task for each client. Each task runs in a loop, attempting to
+    /// start the client and reopening the connection with a new URI if an error occurs.
+    pub async fn start(self) {
```
obelisk, line 148: Is it intended you want this to consume itself? It's not a problem, just making sure that was an intentional decision.

plaid/src/data/websocket/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+        info!(
```
obelisk, line 149: We should out put the names, or some identifying information here

plaid/src/data/websocket/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+    sleep_duration: u64,
+    /// The number of Logbacks this generator is allowed to trigger
+    #[serde(default)]
+    logbacks_allowed: LogbacksAllowed,
+    /// The minimum amount of time (in milliseconds) to wait before retrying a connection to a WebSocket
+    #[serde(default = "min_retry_duration")]
+    min_retry_duration: u64,
+    /// The maximum amount of time (in milliseconds) to wait before retrying a connection to a WebSocket
+    #[serde(default = "max_retry_duration")]
+    max_retry_duration: u64,
```
obelisk, line 54: Is there any reason we don't want to immediately parse these into Durations?

obelisk (changes requested):

plaid/src/data/websocket/mod.rs
wbssbw: Yeah it's intentional. The associated methods on the socket require a mutable reference so I have the start method consume itself

plaid/src/data/websocket/mod.rs
wbssbw: I think the most we can/should do is throw a warning for sockets whose protocol is ws. After that, it's up to whoever is running the instance to accept that risk

plaid/src/data/websocket/selector.rs
```diff
@@ -0,0 +1,124 @@
...
+            initial_retry_after,
+            max_retry_after,
+        }
+    }
+
+    /// Selects the next URI to use, prioritizing URIs with the shortest backoff duration.
+    ///
+    /// This function sorts the URIs by their next attempt time,
+    /// and returns a tuple containing the name and URI of the entry with the shortest backoff duration that is ready for the next attempt.
+    /// If no URIs are ready, it selects and returns the name and URI of the entry with the earliest next attempt time.
```
obelisk, line 77: If this is the case then I think the concept of "Ready" perhaps is not the right word to use. I think we either need to stop the generator until one is ready, or change the wording to be that we will do exponential back off but it won't mean much if they all go down (since we're just going to cycle through them at high speed).

plaid/src/data/websocket/selector.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+/// Manages a list of URI entries and handles the selection and retry logic for connection attempts.
+///
+/// The `UriSelector` struct contains a collection of `UriEntry` instances and manages the logic
+/// for selecting the next URI to attempt a connection to, including backoff and retry mechanisms.
+#[derive(Debug)]
+pub struct UriSelector {
+    /// A list of `UriEntry` instances.
+    uris: Vec<UriEntry>,
```
obelisk, line 30: The data structure to use here is a MinHeap which will keep it sorted for you.

obelisk (changes requested):

plaid/src/data/websocket/mod.rs
```diff
@@ -0,0 +1,466 @@
...
+                    },
+                    _ = read_handle => {
+                        error!("Write task for WebSocket: [{}] using socket [{}] finished unexpectedly", &self.name, uri_name);
+                    },
+                }
+            }
+            None => {
+                tokio::select! {
+                    _ = read_handle => {
+                        error!("Write task for WebSocket: [{}] using socket [{}] finished unexpectedly", &self.name, uri_name);
```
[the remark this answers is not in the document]
wbssbw: Good catch

plaid/src/data/websocket/selector.rs
wbssbw: Good call - implemented waiting until next attempt has passed

plaid/Cargo.toml
```diff
@@ -38,6 +38,8 @@ warp = { version = "0.3", features = ["tls"] }
 wasmer = { version = "4", default-features = false, features = ["cranelift"] }
 wasmer-middlewares = "4"
 jsonwebtoken = { version = "9.2" }
+tokio-tungstenite = { version = "0.23.1", features = ["native-tls"] }
+futures-util = "0.3.30"
```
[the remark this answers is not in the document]
wbssbw: Tokio offers the [`split`](https://docs.rs/tokio/1.39.3/tokio/io/fn.split.html) method that would allow us to split the socket into a read half and a write half. 

Unfortunately, `WebSocketStream<tokio_tungstenite::MaybeTlsStream<tokio::net::TcpStream>>` does not satisfy the `AsyncRead` trait so we are unable to use tokio

plaid/src/data/websocket/mod.rs
```diff
@@ -0,0 +1,466 @@
...
+    /// - `sender`: A channel sender to which log messages are sent.
+    /// - `uri`: The URI of the WebSocket.
+    ///
+    /// # Returns
+    /// A `JoinHandle` for the spawned task.
+    ///
+    /// # Errors
+    /// If an error occurs while reading a message from the WebSocket or sending a log message to the
+    /// channel, the task logs the error and terminates.
+    async fn spawn_read_task(
```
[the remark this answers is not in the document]
wbssbw: Good idea - added it as optional to provide. If None is given, we default to 64kB

plaid/resources/plaid.toml
```diff
@@ -119,6 +119,17 @@ secret_key = ""
...
 
+# [data.websocket]
+# [data.websocket.websockets]
+# [data.websocket."websockets"."demo_rpc_call"]
+# log_type = "testing"
+# [data.websocket."websockets"."demo_rpc_call".uris]
+# simplystaking = "wss://some_websocket"
+# [data.websocket."websockets"."demo_rpc_call".message_config]
+# message = "{ \"id\": 1, \"jsonrpc\": \"2.0\", \"method\": \"eth_getBlockByNumber\", \"params\": [ \"finalized\", false ]}"
+# sleep_duration = 100
```
obelisk, line 130: I would add a second comment here (##) that this means every 0.1 seconds.

plaid/src/data/websocket/selector.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// and returns a tuple containing the name and URI of the entry with the shortest backoff duration that is ready for the next attempt.
+    /// If no URIs are ready, it selects and returns the name and URI of the entry with the earliest next attempt time.
+    ///
+    /// # Returns
+    ///
+    /// A tuple `(String, Uri)` where the `String` is the name of the URI and the `Uri` is the next URI to use.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the URI list is empty, although in practice this should never be possible.
```
obelisk, line 99: Doesn't panic anymore?

plaid/src/data/websocket/selector.rs
```diff
@@ -0,0 +1,141 @@
...
+    pub async fn next_uri(&self) -> Option<(String, Uri)> {
+        // Select the URI with the shortest backoff duration that is ready for the next attempt
+        let uri = self.uris.peek()?;
+        let now = Instant::now();
+
+        if now >= uri.next_attempt {
```
obelisk, line 105: I would flip this so you know if you have to sleep, then you can crush the returns together.

plaid/src/data/websocket/selector.rs
```diff
@@ -84,33 +84,34 @@ impl UriSelector {
...
     pub async fn next_uri(&self) -> Option<(String, Uri)> {
         // Select the URI with the shortest backoff duration that is ready for the next attempt
         let uri = self.uris.peek()?;
         let now = Instant::now();
 
-        if now >= uri.next_attempt {
-            Some((uri.name.clone(), uri.uri.clone()))
-        } else {
+        // If the next attempt hasn't passed, sleep until the socket is ready
+        if uri.next_attempt < now {
```
obelisk, line 109: You subtract now then add it back. Can we not just sleep until `uri.next_attempt`

obelisk (changes requested):

plaid/src/data/websocket/selector.rs
wbssbw: lol excellent point

obelisk (approved):

PR #17, opened 2024-08-26
Plaid 14: Efficient Get and Tests

obelisk:
# Efficient Get
We have had a problem with the GET handling where servicing a GET request required spawning a `pthread` which creates a pretty large attack surface. This fixes that by changing how messages are structured so they now can include an `Arc<PlaidModule>` and a callback channel. This means we can process GET requests in the same message processing system and we can get rid of the `execute_immediate` function.

# Breaking
This PR also fully removes Quorum for several reasons:
1. The functionality it provided can now be fully replicated inside a Plaid module
2. The private dependency made building and testing quite complicated
3. Quorum is generally considered EoL

# New APIs
This adds a new storage API so that existing keys can be listed by the module. It also allows prefix filtering in the host so modules don't have to handle needlessly deserializing the entire key set.

# Refactor
This makes some changes to the internal API, moving storage and cache into submodules. It also changes some types on the `StorageProvider` trait so keys now must be strings. This is something that we were doing anyway due to `safely_get_string` but this makes it more explicit while also bringing the API closer to that of the cache.

## Tests
Since I'm changing such a core part of the Plaid runtime to rework messages and GET request handling, we need tests to ensure we're not breaking anything critical. This works by creating a bunch of test modules which have their own particular test harness called by the new global test harness.


plaid/src/executor/mod.rs
```diff
@@ -279,6 +309,9 @@ fn update_persistent_response(
                 match pr.data.write() {
                     Ok(mut data) => {
                         *data = Some(response);
+                        if let Some(sender) = response_sender {
+                            sender.send(data.clone()).unwrap();
```
wbssbw, line 313: 🗞️🗞️🗞️🗞️🗞️🗞️

plaid/src/bin/plaid.rs
```diff
@@ -227,47 +228,45 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                                        TrySendError::Disconnected(_) => panic!("The execution system is no longer accepting messages. Nothing can continue."),
+                                    }
+                                }
+
+                                match response_recv.await {
+                                    Ok(Some(response))=> {
                                         if update {
                                             info!("Updating cache for get request to: {webhook}");
                                             let mut cache = get_cache.write().await;
                                             cache.insert(webhook.clone(), (current_time, response.clone()));
```
wbssbw, line 258: I would maybe get a new current time before inserting to the cache here. It's possible that enough time passes here such that the cache is already invalid

wbssbw (commented):
Good idea using the oneshot channel

plaid/src/executor/mod.rs
```diff
@@ -19,13 +20,29 @@ use std::collections::HashMap;
...
     pub accessory_data: HashMap<String, Vec<u8>>,
+    /// Where the message came from
     pub source: LogSource,
+    /// If this message is allowed to trigger additional messages to the same
+    /// or other message channels
     pub logbacks_allowed: LogbacksAllowed,
+    /// If a response is should be sent back to the source of the message
+    /// This is used in the GET request system to handle responses
+    #[serde(skip)]
+    pub response_sender: Option<OneShotSender<Option<String>>>,
```
wbssbw, line 40: Could be nice to have this send a status code as well so that we're not always responding with a 200

obelisk: That would require the rule have an API with which to send a response code which it doesn't currently. Should we stub it out to add later or do you feel we should add such an API in this PR?

I'm somewhat against adding it in this PR because it will require rethinking how persistent response works in terms of caching which I'm not sure we need to tackle right now.

plaid/src/executor/mod.rs
obelisk: Definitely 🗞️

plaid/src/bin/plaid.rs
obelisk: That's a good callout

plaid/src/executor/mod.rs
wbssbw: Yeah that's a fair point - let's stub it out as something to add later. We have no use case for needing a status code right now so no reason to add complexity

michelemin (approved):
LGTM 🚀

PR #19, opened 2024-09-03
KMS functionality

wbssbw:
This PR adds (optional) KMS support to the Plaid runtime.

### Methods
We provide users with the `sign_arbitrary_message` method to sign messages with the `get_public_key` method to fetch metadata for a given key ID.

### Configuration
```toml
[apis."aws"]
[apis."aws"."kms".authentication]
[apis."aws"."kms".key_configuration]
"some_key_id" = ["testing_test.wasm", "another_rule.wasm"]
```

This config defines the KMS API using IAM roles for authentication. We also configure 1 key with ID: "some_key_id" and allow the `testing_test.wasm` and `another_rule.wasm` modules to use it.

#### Authentication
We provide two methods of authentication to KMS:
1. `IAM`
3. API keys

When IAM is selected in the configuration, Plaid uses the IAM role assigned to the instance to construct an instance of the KMS client. It is recommended that you use IAM in production environments.

When API keys are used, Plaid use the values provided in the configuration to construct an instance of the KMS client.

### DIFF
--- a/plaid-stl/src/aws/kms.rs
+++ b/plaid-stl/src/aws/kms.rs
@@ -28,6 +28,38 @@ pub enum MessageType {
     Digest,
 }
 
+/// The signing algorithm to use in signing this request
+pub enum SigningAlgorithm {
+    EcdsaSha256,
+    EcdsaSha384,
+    EcdsaSha512,
+    RsassaPkcs1V15Sha256,
+    RsassaPkcs1V15Sha384,
+    RsassaPkcs1V15Sha512,
+    RsassaPssSha256,
+    RsassaPssSha384,
+    RsassaPssSha512,
+}
+
+impl Serialize for SigningAlgorithm {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        match self {
+            Self::EcdsaSha256 => serializer.serialize_str("ECDSA_SHA_256"),
+            Self::EcdsaSha384 => serializer.serialize_str("ECDSA_SHA_384"),
+            Self::EcdsaSha512 => serializer.serialize_str("ECDSA_SHA_512"),
+            Self::RsassaPkcs1V15Sha256 => serializer.serialize_str("RSASSA_PKCS1_V1_5_SHA_256"),
+            Self::RsassaPkcs1V15Sha384 => serializer.serialize_str("RSASSA_PKCS1_V1_5_SHA_384"),
+            Self::RsassaPkcs1V15Sha512 => serializer.serialize_str("RSASSA_PKCS1_V1_5_SHA_512"),
+            Self::RsassaPssSha256 => serializer.serialize_str("RSASSA_PSS_SHA_256"),
+            Self::RsassaPssSha384 => serializer.serialize_str("RSASSA_PSS_SHA_384"),
+            Self::RsassaPssSha512 => serializer.serialize_str("RSASSA_PSS_SHA_512"),
+        }
+    }
+}
+
 /// Creates and sends a named signing request to the KMS API.
 ///
 /// This function:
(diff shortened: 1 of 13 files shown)


plaid-stl/src/aws/kms.rs
```diff
@@ -0,0 +1,193 @@
...
+/// Tells KMS whether the value of the Message parameter should be hashed as part of the signing algorithm.
+/// Use `Raw` for unhashed messages; use `Digest` for message digests, which are already hashed.
+#[derive(Serialize)]
+pub enum MessageType {
+    Raw,
+    Digest,
+}
+
+/// The signing algorithm to use in signing this request
+pub enum SigningAlgorithm {
```
obelisk, line 46: Is there nothing that distinguishes the K1 and R1 curves?

plaid/resources/plaid.toml
```diff
@@ -117,6 +117,25 @@ allowed_rules = ["testing_test.wasm"]
 client_id = 99999
 secret_key = ""
 
+# KMS USING IAM
+
+# [apis."aws"]
+# [apis."aws"."kms".authentication]
+# [apis."aws"."kms".key_configuration]
```
obelisk, line 124: If authentication is empty, IDMS/IAM just works TM?

plaid/src/bin/plaid.rs
```diff
@@ -53,7 +53,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     info!("Configurating APIs for Modules");
     // Create the API that powers all the wrapped calls that modules can make
-    let api = Api::new(config.apis, log_sender.clone(), delayed_log_sender);
+    let api = Api::new(config.apis, log_sender.clone(), delayed_log_sender).await;
```
obelisk, line 56: Ah, this is a pretty major change. I think it's fine but have we considered any other places this could be an issue?

plaid/src/functions/api.rs
```diff
@@ -130,6 +130,95 @@ macro_rules! impl_new_function_with_error_buffer {
...
+/// - `$sub_module`: The name of the submodule within the API (e.g., `kms`).
+/// - `$function_name`: The name of the function to be implemented (e.g., `put_object`, `encrypt`).
+///
+/// # Error Handling
+/// The generated implementation function returns `FunctionErrors` in case of failures, which are then
+/// converted to int error codes by the wrapper function. These errors include:
+/// - `FunctionErrors::InternalApiError`: For internal API-related errors.
+/// - `FunctionErrors::ApiNotConfigured`: If the API is not configured.
+/// - `FunctionErrors::ReturnBufferTooSmall`: If the provided return buffer is too small to hold the result.
+macro_rules! impl_new_sub_module_function_with_error_buffer {
```
obelisk, line 153: I think to crush the two very similar macros together you need to modify this so it expects the submodule to have the trailing `.`.

Then you have an interim function with this name `impl_new_sub_module_function_with_error_buffer` which add the final `.` to the submodule before calling this.

And the existing one can call it directly because there is no submodule. Do you think that may work?

plaid/src/bin/plaid.rs
wbssbw: Yeah I wanted to avoid this if possible but unfortunately has to happen. I can't think of any scenario where this would be ready before a rule ended up needing it. We could always block on this if we're concerned

plaid/src/functions/api.rs
wbssbw: Worth trying!

plaid/resources/plaid.toml
wbssbw: Good question - I'll test and let you know

plaid-stl/src/aws/kms.rs
wbssbw: Not that I'm aware of - here is the AWS docs:
https://docs.rs/aws-sdk-kms/1.42.0/aws_sdk_kms/types/enum.SigningAlgorithmSpec.html

plaid/resources/plaid.toml
wbssbw: Yeah it'll default to IAM if no keys are provided

plaid/src/bin/plaid.rs
obelisk: I think it's fine

obelisk (approved):

PR #20, opened 2024-09-04
Support paginated responses for GH get_repository_collaborators

michelemin:
This PR adds support for paginated responses from the GitHub API when fetching a repo's collaborators. See [the docs](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#list-repository-collaborators) for more details about that endpoint.

The main idea is that we will always make a GET call with query params `page` and `per_page`. If these params are not given, then we default to 1 and 30, respectively (which are the defaults used by GH).

The existing `get_repository_collaborators` in plaid-stl has been marked as deprecated, and `get_all_repository_collaborators` has been added.

## Backwards compatibility
The changes are backwards compatible because
* In the plaid-stl we only add `get_all_repository_collaborators`
* In the plaid API, we expect extra parameters `page` and `per_page`, but fallback on default values if those are missing

### DIFF
--- a/plaid-stl/src/github/mod.rs
+++ b/plaid-stl/src/github/mod.rs
@@ -1,12 +1,32 @@
 use std::{collections::HashMap, fmt::Display};
 
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 
 use crate::PlaidFunctionError;
 
 pub enum ReviewPatAction {
     Approve,
     Deny,
+}
+
+#[derive(Debug, Deserialize)]
+/// Set of user permissions, as returned by GitHub's API
+pub struct Permission {
+    pub pull: bool,
+    pub triage: bool,
+    pub push: bool,
+    pub maintain: bool,
+    pub admin: bool,
+}
+
+#[derive(Debug, Deserialize)]
+/// A collaborator on a GitHub repository
+pub struct RepositoryCollaborator {
+    pub login: String,
+    #[serde(rename = "type")]
+    pub type_: String,
+    pub role_name: String,
+    pub permissions: Permission,
 }
 
 // TODO: Do not use this function, it is deprecated and will be removed soon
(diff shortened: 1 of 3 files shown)


obelisk:
Major comment would be to deserialize the data into a Vec<GitHubUser> or similar which will make it much easier for the consumer.

plaid/resources/plaid.toml
```diff
@@ -33,7 +33,7 @@ okta = 9_000_000
 "example_rule.wasm" = 5_000_000
 
 [loading.memory_page_count]
-default = 250
+default = 300
```
michelemin, line 36: Tested with a repository which had 361 collaborators, and I was hitting the memory limit.

plaid-stl/src/github/mod.rs
```diff
@@ -9,6 +9,25 @@ pub enum ReviewPatAction {
...
+    push: bool,
+    maintain: bool,
+    admin: bool,
+}
+
+#[derive(Debug, Deserialize)]
+/// A collaborator on a GitHub repository
+pub struct RepositoryCollaborator {
+    login: String,
+    r#type: String,
```
obelisk, line 26: Generally I prefer `type_` and when used with serde you can use a rename directive. That just means that we prevent having `r#` throughout the codebase.

Also I think these variables need to be marked `pub` so they can be accessed by calling rules.

plaid-stl/src/github/mod.rs
```diff
@@ -9,6 +9,25 @@ pub enum ReviewPatAction {
     Deny,
 }
 
+#[derive(Debug, Deserialize)]
+/// Set of user permissions, as returned by GitHub's API
+struct Permission {
```
obelisk, line 14: I think this also needs to be pub

obelisk (changes requested):
Can you also clarify the backwards compatibility of this? I don't care if it's not backwards compatible, just want to know if rules need to be recompiled or not

michelemin:
> Can you also clarify the backwards compatibility of this? I don't care if it's not backwards compatible, just want to know if rules need to be recompiled or not

Added in the PR's comment

obelisk (approved):

PR #21, opened 2024-09-09
KMS tweaks

wbssbw:
Addresses bugs I've run into when trying to write a rule using the new KMS APIs.

Passing the message from KMS to sign as a `String` was silly of me - have made that a `Vec<u8>` now which allows us to do away with the `parse_message` function. 

I also ran into issues using API key auth with short lived credentials to AWS as I was missing a session token. We now allow for an optional session token in the `Authentication` enum.

Finally, I linked the functions incorrectly and have now fixed that

(Also enables AWS by default)

### DIFF
--- a/plaid/src/apis/aws/kms.rs
+++ b/plaid/src/apis/aws/kms.rs
@@ -22,8 +22,7 @@ struct SignRequestRequest {
     /// - Alias ARN: arn:aws:kms:us-east-2:111122223333:alias/ExampleAlias
     key_id: String,
     /// Specifies the message or message digest to sign. Messages can be 0-4096 bytes. To sign a larger message, provide a message digest.
-    #[serde(deserialize_with = "parse_message")]
-    message: Blob,
+    message: Vec<u8>,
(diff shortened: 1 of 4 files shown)


plaid/src/apis/aws/kms.rs
```diff
@@ -79,6 +68,7 @@ enum Authentication {
     ApiKey {
         access_key_id: String,
         secret_access_key: String,
+        session_token: Option<String>,
```
obelisk, line 71: What is this?

plaid/src/functions/api.rs
```diff
@@ -387,12 +387,12 @@ pub fn to_api_function(
 
         // KMS calls
         #[cfg(feature = "aws")]
-        "aws_kms_sign_arbitrary_message" => {
+        "kms_sign_arbitrary_message" => {
```
obelisk, line 390: Why the name change?

obelisk (approved):

obelisk (approved):

PR #24, opened 2024-09-23
Add npm API to Plaid

michelemin:
### Supported functionality
1. Publish empty package stub
2. Invite user to organization
3. Remove user from organization
4. Add user to team
5. Remove user from team
6. Get list of users in the organization
7. Get list of users in the organization without 2FA enabled
8. Set team permissions on a package
9. Create granular token for package
10. List all granular tokens (with expiration date)
11. Delete a package (TBC if we want to expose this)
12. List packages over which a team has certain permissions (useful to spot misconfigured packages)

### Design choices
* Shared structs are all defined in the STL, since `plaid` depends on `plaid-stl`. Therefore, the code can serialize and deserialize using the same types across the host/guest boundary.
* The TOTP functionality is embedded in the code that uses it. TBD if we want to separate it out.
* The generation of a tarball (for uploading a new package stub) is currently part of the plaid runtime.

### Backward compatibility
Changes are purely additive, so backward compatibility is maintained.

### New configuration for npm API

```
[apis.npm]
username = ""
password = ""
otp_secret = ""
automation_token = ""
npm_scope = ""
user_agent = ""
```

### DIFF
--- a/plaid-stl/src/npm/mod.rs
+++ b/plaid-stl/src/npm/mod.rs
@@ -7,7 +7,7 @@ use chrono::{DateTime, Utc};
 // All the structs which are also used in the Plaid API
 
 /// Permission that is granted to a team over an npm package
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Clone)]
 pub enum NpmPackagePermission {
     READ,
     WRITE,
@@ -18,6 +18,19 @@ impl ToString for NpmPackagePermission {
         match self {
             Self::READ => "read".to_string(),
             Self::WRITE => "write".to_string(),
+        }
+    }
+}
+
+impl NpmPackagePermission {
+    // We do not implement the TryFrom trait because of the error type. We do not want to carry the
+    // NpmError struct here in the STL. Also, we don't care about which error we get from here. If
+    // it fails, the caller will know what to do.
+    pub fn try_from(value: String) -> Result<Self, ()> {
+        match value.as_str() {
+            "read" => Ok(NpmPackagePermission::READ),
+            "write" => Ok(NpmPackagePermission::WRITE),
+            _ => Err(()),
         }
     }
 }
@@ -99,7 +112,7 @@ impl ToString for GranularTokenPackagesAndScopesPermission {
     }
 }
 
-/// TODO I don't exactly know what this means
+/// TODO I don't know what this means
 #[derive(Clone, Serialize, Deserialize)]
 pub enum GranularTokenSelectedPackagesAndScopes {
     PackagesAndScopesSome,
@@ -225,6 +238,12 @@ impl PkgAccessLevel {
 pub struct PublishEmptyStubParams {
     pub package_name: String,
     pub access_level: PkgAccessLevel,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct ListPackagesWithTeamPermissionParams {
+    pub team: NpmTeam,
+    pub permission: NpmPackagePermission,
 }
 
 // End of "All the structs which are also used in the Plaid API"
@@ -537,4 +556,52 @@ pub fn delete_package(package_name: &str) -> Result<(), PlaidFunctionError> {
     }
 
     Ok(())
+}
+
+/// Return a list of npm packages over which a team has a certain permission (read or write)
+pub fn list_packages_with_team_permission(
+    team: NpmTeam,
+    permission: NpmPackagePermission,
+) -> Result<Vec<String>, PlaidFunctionError> {
+    extern "C" {
+        new_host_function_with_error_buffer!(npm, list_packages_with_team_permission);
+    }
+
+    let params = serde_json::to_string(&ListPackagesWithTeamPermissionParams { team, permission })
+        .map_err(|_| PlaidFunctionError::ErrorCouldNotSerialize)?;
+
+    const RETURN_BUFFER_SIZE: usize = 1024 * 1024; // 1 MiB
+    let mut return_buffer = vec![0; RETURN_BUFFER_SIZE];
+
+    let res = unsafe {
+        npm_list_packages_with_team_permission(
+            params.as_bytes().as_ptr(),
+            params.as_bytes().len(),
+            return_buffer.as_mut_ptr(),
+            RETURN_BUFFER_SIZE,
+        )
+    };
+
+    if res < 0 {
+        return Err(res.into());
+    }
+
+    return_buffer.truncate(res as usize);
+    // This should be safe because unless the Plaid runtime is expressly trying
+    // to mess with us, this came from a String in the API module.
+    let res = String::from_utf8(return_buffer).unwrap();
+
+    // The response that comes back contains package name and permission.
+    // However, we only care about the name so we keep only that.
+    #[derive(Deserialize)]
+    struct Package {
+        package_name: String,
+    }
+
+    let v: Vec<String> = serde_json::from_str::<Vec<Package>>(&res)
+        .map_err(|_| PlaidFunctionError::InternalApiError)?
+        .iter()
+        .map(|v| v.package_name.clone())
+        .collect();
+    Ok(v)
 }
--- a/plaid-stl/src/npm/mod.rs
+++ b/plaid-stl/src/npm/mod.rs
@@ -1,0 +1,540 @@
+use serde::{Deserialize, Deserializer, Serialize};
+
+use crate::PlaidFunctionError;
+
+use chrono::{DateTime, Utc};
+
+// All the structs which are also used in the Plaid API
+
+/// Permission that is granted to a team over an npm package
+#[derive(Serialize, Deserialize)]
+pub enum NpmPackagePermission {
+    READ,
+    WRITE,
+}
+
+impl ToString for NpmPackagePermission {
+    fn to_string(&self) -> String {
+        match self {
+            Self::READ => "read".to_string(),
+            Self::WRITE => "write".to_string(),
+        }
+    }
+}
+
+/// Teams in the npm organization
+#[derive(Serialize, Deserialize)]
+pub enum NpmTeam {
+    Admins,
+    Developers,
+}
+
+impl ToString for NpmTeam {
+    fn to_string(&self) -> String {
+        match self {
+            Self::Admins => "Admins".to_string(),
+            Self::Developers => "developers".to_string(),
+        }
+    }
+}
+
+/// Role that a user can have in the npm organization
+#[derive(Debug, Deserialize, Serialize)]
+pub enum NpmUserRole {
+    Member,
+    Admin,
+    Owner,
+}
+
+impl NpmUserRole {
+    // We do not implement the TryFrom trait because of the error type. We do not want to carry the
+    // NpmError struct here in the STL. Also, we don't care about which error we get from here. If
+    // it fails, the caller will know what to do.
+    pub fn try_from(value: String) -> Result<Self, ()> {
+        match value.as_str() {
+            "developer" => Ok(NpmUserRole::Member),
+            "admin" => Ok(NpmUserRole::Admin), // untested: I think we do not have Admins
+            "super-admin" => Ok(NpmUserRole::Owner),
+            _ => Err(()),
+        }
+    }
+}
+
+/// A user in the npm organization
+#[derive(Debug, Serialize, Deserialize)]
+pub struct NpmUser {
+    pub username: String,
+    pub role: NpmUserRole,
+}
+
+/// Org-level permission granted to a granular token.
+///
+/// Note: this is incomplete. Other permission levels are possible, but we never needed to add them.
+#[derive(Clone, Serialize, Deserialize)]
+pub enum GranularTokenOrgsPermission {
+    NoAccess,
+}
+
+impl ToString for GranularTokenOrgsPermission {
+    fn to_string(&self) -> String {
+        match self {
+            GranularTokenOrgsPermission::NoAccess => "No access".to_string(),
+        }
+    }
+}
+
+/// Package-level permission granted to a granular token.
+///
+/// Note: this is incomplete. Other permission levels are possible, but we never needed to add them.
+#[derive(Clone, Serialize, Deserialize)]
+pub enum GranularTokenPackagesAndScopesPermission {
+    ReadAndWrite,
+}
+
+impl ToString for GranularTokenPackagesAndScopesPermission {
+    fn to_string(&self) -> String {
+        match self {
+            GranularTokenPackagesAndScopesPermission::ReadAndWrite => "Read and write".to_string(),
+        }
+    }
+}
+
+/// TODO I don't exactly know what this means
+#[derive(Clone, Serialize, Deserialize)]
+pub enum GranularTokenSelectedPackagesAndScopes {
+    PackagesAndScopesSome,
+}
+
+impl ToString for GranularTokenSelectedPackagesAndScopes {
+    fn to_string(&self) -> String {
+        match self {
+            GranularTokenSelectedPackagesAndScopes::PackagesAndScopesSome => {
+                "packagesAndScopesSome".to_string()
+            }
+        }
+    }
+}
+
+/// Groups together all the input that we expect from a rule that requests
+/// the generation of a granular token. Optional fields will have a default value.
+#[derive(Serialize, Deserialize)]
+pub struct GranularTokenSpecs {
+    pub allowed_ip_ranges: Option<Vec<String>>,
+    pub expiration_days: Option<u16>,
+    pub orgs_permission: Option<GranularTokenOrgsPermission>,
+    pub packages_and_scopes_permission: Option<GranularTokenPackagesAndScopesPermission>,
+    pub selected_orgs: Option<Vec<String>>,
+    pub selected_packages: Option<Vec<String>>,
+    pub selected_packages_and_scopes: Option<GranularTokenSelectedPackagesAndScopes>,
+    pub selected_scopes: Option<Vec<String>>,
+    pub token_name: String,
+    pub token_description: String,
+}
+
+impl GranularTokenSpecs {
+    /// Provide a name and description for the token, and set all other fields to None, which
+    /// will result in default values being used.
+    pub fn with_name_and_description(token_name: &str, token_description: &str) -> Self {
+        // TODO We could check some things about name and description. E.g., the description
+        // should be longer than 8 characters to avoid "test" and "testing".
+        Self {
+            token_name: token_name.to_string(),
+            token_description: token_description.to_string(),
+            allowed_ip_ranges: None,
+            expiration_days: None,
+            orgs_permission: None,
+            packages_and_scopes_permission: None,
+            selected_orgs: None,
+            selected_packages: None,
+            selected_packages_and_scopes: None,
+            selected_scopes: None,
+        }
+    }
+}
+
+/// An npm token configured on the npm website for the current user
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[allow(dead_code)]
+pub struct NpmToken {
+    token: String,
+    token_name: Option<String>,
+    pub token_type: Option<String>,
+    // Deserialize the date field from the ISO 8601 format
+    #[serde(default, deserialize_with = "deserialize_option_timestamp")]
+    expires: Option<DateTime<Utc>>,
+}
+
+// Custom deserializer for an optional DateTime<Utc>
+fn deserialize_option_timestamp<'de, D>(deserializer: D) -> Result<Option<DateTime<Utc>>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    // Try to deserialize as a string (ISO 8601 format) or return None if missing
+    let opt = Option::<String>::deserialize(deserializer)?;
+
+    // Parse the timestamp string into a DateTime<Utc>, if it's present
+    Ok(opt.and_then(|s| {
+        DateTime::parse_from_rfc3339(&s)
+            .ok()
+            .map(|dt| dt.with_timezone(&Utc))
+    }))
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct SetTeamPermissionOnPackageParams {
+    pub team: String,
+    pub package: String,
+    pub permission: NpmPackagePermission,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct CreateGranularTokenForPackageParams {
+    pub package: String,
+    pub specs: GranularTokenSpecs,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct AddRemoveUserToFromTeamParams {
+    pub user: String,
+    pub team: NpmTeam,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct InviteUserToOrganizationParams {
+    pub user: String,
+    pub team: Option<NpmTeam>,
+}
+
+/// Access level for an npm package
+#[derive(Serialize, Deserialize)]
+pub enum PkgAccessLevel {
+    Restricted,
+    Public,
+}
+
+impl PkgAccessLevel {
+    pub fn to_string(&self) -> String {
+        match self {
+            PkgAccessLevel::Restricted => "restricted".to_string(),
+            PkgAccessLevel::Public => "public".to_string(),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct PublishEmptyStubParams {
+    pub package_name: String,
+    pub access_level: PkgAccessLevel,
+}
+
+// End of "All the structs which are also used in the Plaid API"
+
+/// Retrieve a list of users in the npm organization
+pub fn get_org_user_list() -> Result<Vec<NpmUser>, PlaidFunctionError> {
+    extern "C" {
+        new_host_function_with_error_buffer!(npm, get_org_user_list);
+    }
+    const RETURN_BUFFER_SIZE: usize = 32 * 1024; // 32 KiB
+    let mut return_buffer = vec![0; RETURN_BUFFER_SIZE];
+
+    let params = "".to_string();
+
+    let res = unsafe {
+        npm_get_org_user_list(
+            params.as_bytes().as_ptr(),
+            params.as_bytes().len(),
+            return_buffer.as_mut_ptr(),
+            RETURN_BUFFER_SIZE,
+        )
+    };
+
+    if res < 0 {
+        return Err(res.into());
+    }
+
+    return_buffer.truncate(res as usize);
+    // This should be safe because unless the Plaid runtime is expressly trying
+    // to mess with us, this came from a String in the API module.
+    let res = String::from_utf8(return_buffer).unwrap();
+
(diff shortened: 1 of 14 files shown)


plaid/src/apis/npm/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+use regex::Regex;
+use reqwest::{cookie::Jar, Client};
+use serde::{Deserialize, Serialize};
+
+/// Client for interacting with npm that groups a client for CLI operations
+/// and one for web operations
+pub struct Npm {
+    config: NpmConfig,
+    client: Client,
+    cookie_jar: Arc<Jar>,
```
michelemin, line 16: @obelisk TBC if we need a `Mutex` here

plaid-stl/src/npm/mod.rs
```diff
@@ -0,0 +1,607 @@
...
+// All the structs which are also used in the Plaid API
+
+/// Permission that is granted to a team over an npm package
+#[derive(Serialize, Deserialize, Clone)]
+pub enum NpmPackagePermission {
+    READ,
+    WRITE,
+}
+
+impl ToString for NpmPackagePermission {
```
obelisk, line 16: I think this should be `impl Display`?

plaid-stl/src/npm/mod.rs
```diff
@@ -0,0 +1,607 @@
...
+    fn to_string(&self) -> String {
+        match self {
+            Self::READ => "read".to_string(),
+            Self::WRITE => "write".to_string(),
+        }
+    }
+}
+
+impl NpmPackagePermission {
+    // We do not implement the TryFrom trait because of the error type. We do not want to carry the
```
obelisk, line 26: Not sure I follow

plaid-stl/src/npm/mod.rs
```diff
@@ -0,0 +1,607 @@
...
+            "read" => Ok(NpmPackagePermission::READ),
+            "write" => Ok(NpmPackagePermission::WRITE),
+            _ => Err(()),
+        }
+    }
+}
+
+/// Teams in the npm organization
+#[derive(Serialize, Deserialize)]
+pub enum NpmTeam {
```
obelisk, line 40: Is this defined by NPM? Can't we have arbitrary teams?

plaid-stl/src/npm/mod.rs
```diff
@@ -0,0 +1,607 @@
...
+    Admins,
+    Developers,
+}
+
+impl ToString for NpmTeam {
+    fn to_string(&self) -> String {
+        match self {
+            Self::Admins => "Admins".to_string(),
```
obelisk, line 48: One is capital the other is not

plaid-stl/src/npm/mod.rs
```diff
@@ -0,0 +1,607 @@
...
+
+impl NpmUserRole {
+    // We do not implement the TryFrom trait because of the error type. We do not want to carry the
+    // NpmError struct here in the STL. Also, we don't care about which error we get from here. If
+    // it fails, the caller will know what to do.
+    pub fn try_from(value: String) -> Result<Self, ()> {
+        match value.as_str() {
+            "developer" => Ok(NpmUserRole::Member),
+            "admin" => Ok(NpmUserRole::Admin), // untested: I think we do not have Admins
+            "super-admin" => Ok(NpmUserRole::Owner),
```
obelisk, line 70: Is this NPMness of sometimes calling it super-admin, and other times owner?

plaid-stl/src/npm/mod.rs
```diff
@@ -0,0 +1,607 @@
...
+/// A user in the npm organization
+#[derive(Debug, Serialize, Deserialize)]
+pub struct NpmUser {
+    pub username: String,
+    pub role: NpmUserRole,
+}
+
+/// Org-level permission granted to a granular token.
+///
+/// Note: this is incomplete. Other permission levels are possible, but we never needed to add them.
```
obelisk, line 85: I'd add the other ones here anyway just for completeness of the API

plaid-stl/src/npm/mod.rs
```diff
@@ -0,0 +1,607 @@
...
+        match self {
+            GranularTokenOrgsPermission::NoAccess => "No access".to_string(),
+        }
+    }
+}
+
+/// Package-level permission granted to a granular token.
+///
+/// Note: this is incomplete. Other permission levels are possible, but we never needed to add them.
+#[derive(Clone, Serialize, Deserialize)]
```
obelisk, line 102: Again, would complete this to make it easier to add things later without having to change or update the runtime

plaid-stl/src/npm/mod.rs
```diff
@@ -0,0 +1,607 @@
...
+
+impl ToString for GranularTokenPackagesAndScopesPermission {
+    fn to_string(&self) -> String {
+        match self {
+            GranularTokenPackagesAndScopesPermission::ReadAndWrite => "Read and write".to_string(),
+        }
+    }
+}
+
+/// TODO I don't know what this means
```
obelisk, line 115: Maybe Andrew can help

plaid-stl/src/npm/mod.rs
```diff
@@ -0,0 +1,607 @@
...
+}
+
+/// Access level for an npm package
+#[derive(Serialize, Deserialize)]
+pub enum PkgAccessLevel {
+    Restricted,
+    Public,
+}
+
+impl PkgAccessLevel {
```
obelisk, line 228: `impl Display for`

plaid-stl/src/npm/mod.rs
```diff
@@ -0,0 +1,607 @@
...
+    pub access_level: PkgAccessLevel,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct ListPackagesWithTeamPermissionParams {
+    pub team: NpmTeam,
+    pub permission: NpmPackagePermission,
+}
+
+// End of "All the structs which are also used in the Plaid API"
```
obelisk, line 249: If this is a meaningful distinction, I would move them to another file so they can be in their own mod. That'll also make it a lot more clear when updating

plaid-stl/src/npm/mod.rs
```diff
@@ -0,0 +1,607 @@
...
+    // to mess with us, this came from a String in the API module.
+    let res = String::from_utf8(return_buffer).unwrap();
+
+    serde_json::from_str::<Vec<NpmUser>>(&res).map_err(|_| PlaidFunctionError::InternalApiError)
+}
+
+/// Invite a user to join the npm organization. If the user accepts the invite, they will be added
+/// to the default team "developers".
+///
+/// TODO The Plaid API supports specifying another team as the default for the new user. Expose through the STL?
```
obelisk, line 316: TODO in code.

Not sure how to understand the question though.

plaid-stl/src/npm/mod.rs
```diff
@@ -0,0 +1,607 @@
+use serde::{Deserialize, Deserializer, Serialize};
```
obelisk, line 1: This whole file is calling out for some macros to remove a lot of the boilerplate. Might be something to consider doing here then applying it to the rest of the STL.

plaid/src/apis/npm/mod.rs
obelisk: If it works without it then I think ti's fine. Rust should prevent you from doing something dangerous so this item must have interior mutability.

plaid/src/apis/npm/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+/// Credentials and secrets for interacting with npm
+pub struct NpmConfig {
+    /// Username for the npm account
+    pub username: String,
+    /// Password for the npm account
+    pub password: String,
+    /// Secret for the TOTP-based 2FA on the npm account
+    pub otp_secret: String,
```
obelisk, line 42: This should probably be optional because not all orgs enforce this

plaid/src/apis/npm/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        // Check the automation_token looks OK: it should be "npm_" followed by 36 alphanum characters
+        let automation_token_regex = Regex::new(r"^npm_[a-zA-Z0-9]{36}$").unwrap();
+        if !automation_token_regex.is_match(&automation_token) {
+            return Err(NpmError::WrongConfig(
+                "Wrong format for automation token".to_string(),
+            ));
+        }
+
+        // TODO Quickly try to do a login and see if the credentials work? Otherwise we will discover
```
obelisk, line 79: TODO in code

plaid/src/apis/npm/npm_cli_client.rs
```diff
@@ -0,0 +1,191 @@
...
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use tar::{Builder, Header};
+
+const REGISTRY_URL: &str = "https://registry.npmjs.org";
+const NODE_VERSION: &str = "20.16.0";
+const NPM_VERSION: &str = "10.8.3";
+
+#[derive(Serialize)]
+struct PkgAttachment<'a> {
```
obelisk, line 17: This structures are going to need some doc comments to explain what they are all for otherwise it's going to be very hard for someone else to come in an add to it.

plaid/src/apis/npm/npm_cli_client.rs
```diff
@@ -0,0 +1,191 @@
...
+
+#[derive(Serialize)]
+struct PkgVersion<'a> {
+    name: &'a str,
+    version: &'a str,
+    main: &'a str,
+    author: &'a str,
+    license: &'a str,
+    _id: &'a str,
+    readme: &'a str,
```
obelisk, line 38: Is this a reference to a readme, or the actual readme? if it's the actual readme, I'm concerned about the reallocing here.

plaid/src/apis/npm/npm_cli_client.rs
```diff
@@ -0,0 +1,191 @@
...
+        info!("Publishing new empty npm package [{}] on behalf of [{module}]", params.package_name);
+        
+        let (package_json, tarball_data) =
+            create_package_tarball(&self.config.npm_scope, &params.package_name)?;
+        let data_length = tarball_data.len();
+        let sha1_digest = hashes::sha1_hex(&tarball_data);
+        let sha512_digest = hashes::sha512_base64(&tarball_data);
+
+        // Safe unwrap: we are hardcoding the json content, so the deserialization will not fail
+        let manifest = serde_json::from_str::<PkgManifest>(&package_json).unwrap();
```
obelisk, line 83: We still should not unwrap here, we can easily capture and return the error. This just seems like asking for trouble.

plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -0,0 +1,827 @@
...
+            serde_json::from_str(params).map_err(|_| ApiError::BadRequest)?;
+        let team = params.team.unwrap_or(plaid_stl::npm::NpmTeam::Developers);
+
+        info!(
+            "Inviting user [{}] to npm organization and team [{}] on behalf of [{module}]",
+            params.user,
+            team.to_string()
+        );
+
+        let body = format!(
```
obelisk, line 494: This will allow injection. You should create a struct and serialize it from there

plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -0,0 +1,827 @@
...
+            .await
+            .map_err(|_| ApiError::NpmError(NpmError::LoginFlowError))?;
+
+        let response: Value = self
+            .client
+            .get(format!(
+                "{}/settings/{}/teams/team/{}/access",
+                NPMJS_COM_URL,
+                self.config.npm_scope,
+                params.team.to_string()
```
obelisk, line 641: This is another injection possibility for rules. This needs to be sanitized.

plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -0,0 +1,827 @@
...
+
+        // Get all packages the team can access and keep only those
+        // that have the desired permission (specified in the params)
+        let packages = self
+            .get_paginated_data_from_npm_website::<NpmPackageWithPermission>(
+                &format!(
+                    "{}/settings/{}/teams/team/{}/access",
+                    NPMJS_COM_URL,
+                    self.config.npm_scope,
+                    params.team.to_string()
```
obelisk, line 670: Similar.

obelisk (changes requested):

plaid-stl/src/npm/mod.rs
michelemin: This is replicating what we already have configured in the organization. In theory, one can always create new teams from the website and add them here, if they should be managed by the API.
For example, on the website there are two more teams for auditors that I did not expose here.

plaid-stl/src/npm/mod.rs
michelemin: Yep :)
But these are the names that are configured on the website, so we need those to appear in the requests we make

plaid-stl/src/npm/mod.rs
michelemin: Yes, as far as I can tell, these terms are used interchangeably in different parts of the API / website

plaid/src/apis/npm/npm_cli_client.rs
michelemin: In reality, all this is used to create a single request, for initially publishing an empty stub. In that sense, we have complete control over each of these fields, and they are not exposed to rules. To be discussed if we can keep it this way.

plaid-stl/src/npm/mod.rs
michelemin: When inviting a user into an organization, the default is to add the user to the `developers` team. However, npm (and the API) support inviting a user _directly_ into another team, like `Admins`, instead of inviting and then promoting.
We can expose this through the STL and take it as a parameter, if we think there is no risk of abuse or error from the rule's part.

plaid-stl/src/npm/mod.rs
michelemin: Please ignore :)
I have now implemented the proper `TryFrom`

plaid/src/apis/npm/mod.rs
michelemin: Good point, done

plaid/src/apis/npm/mod.rs
michelemin: Right, wanted to discuss if this is a good idea or not. Andrew suggested it's not needed because the login happens very early on in every flow, so a failure would appear immediately.

plaid/src/apis/npm/npm_cli_client.rs
michelemin: Fixed

plaid-stl/src/npm/mod.rs
michelemin: If we want to make everything generic, and indeed have arbitrary teams, we can just take this as a rule-defined string.

The idea of using an enum was to prevent silly mistakes like typos, but the flip side is that we have to know all the possibilities beforehand.

michelemin: Moved to using strings so that rules have more flexibility. Let's see if this works fine.

plaid-stl/src/npm/mod.rs
michelemin: Removed when moving to using strings instead of enum for teams.

plaid-stl/src/npm/mod.rs
michelemin: Good idea, moved to separate module in the STL.

plaid/src/apis/npm/npm_web_client.rs
michelemin: Right. Sanitized user name and team name

plaid/src/apis/npm/npm_cli_client.rs
michelemin: Added a comment to explain that these structs are just to mimic the behavior of `npm publish` and should not be touched.

plaid-stl/src/npm/mod.rs
michelemin: Figured out

plaid-stl/src/npm/mod.rs
```diff
@@ -312,16 +69,17 @@ pub fn get_org_users_without_2fa() -> Result<Vec<NpmUser>, PlaidFunctionError> {
 
 /// Invite a user to join the npm organization. If the user accepts the invite, they will be added
 /// to the default team "developers".
-///
-/// TODO The Plaid API supports specifying another team as the default for the new user. Expose through the STL?
-pub fn invite_user_to_organization(user: &str) -> Result<(), PlaidFunctionError> {
+pub fn invite_user_to_organization(
+    user: &str,
```
obelisk, line 73: You might consider replacing these `&str`, with `impl Display`

plaid-stl/src/npm/shared_structs/mod.rs
```diff
@@ -0,0 +1,238 @@
+//! This module contains all structs which are shared between the STL and the runtime API.
```
obelisk, line 1: What's the semantic meaning of `//!`

plaid/src/apis/npm/mod.rs
```diff
@@ -22,11 +25,18 @@ impl Npm {
         let client = Client::builder()
             .cookie_provider(cookie_jar.clone())
             .build()
-            .map_err(|_| NpmError::GenericError).unwrap(); // TODO @obelisk OK to unwrap here? If we cannot build the client, perhaps we should just panic?
+            .map_err(|_| NpmError::GenericError)
+            .unwrap(); // TODO @obelisk OK to unwrap here? If we cannot build the client, perhaps we should just panic?
```
obelisk, line 29: This comes with a philosophical question about Plaid: if an API is misconfigured or cannot start, what should be done. Based on the fact that you have `NpmError::GenericError` we *could* handle it. My preference would be to make the error more descriptive and then return control back to the higher level to decide if API configuration failure are something it wants to allow (that means in the future we can make the decision in one place instead of in every module) but I don't feel so strongly about it as to block the PR on it.

plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -107,35 +108,43 @@ impl Npm {
             ])
             .send()
             .await
+            .map(|_| ())
```
obelisk, line 111: I assume that this means there is no value in the response which could be useful to the caller. If so, just resolve this comment.

plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -406,11 +416,12 @@ impl Npm {
...
         let params: AddRemoveUserToFromTeamParams =
             serde_json::from_str(params).map_err(|_| ApiError::BadRequest)?;
+        let user = self.validate_npm_username(&params.user)?;
+        let team = self.validate_npm_team_name(&params.team)?;
 
         info!(
             "Removing user [{}] from team [{}] on behalf of [{module}]",
-            params.user,
-            params.team.to_string()
+            user, team
```
obelisk, line 424: You can use `{user}` and `{team}` syntax to clean this up a little

plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -483,12 +491,13 @@ impl Npm {
             .map_err(|_| ApiError::NpmError(NpmError::LoginFlowError))?;
         let params: InviteUserToOrganizationParams =
             serde_json::from_str(params).map_err(|_| ApiError::BadRequest)?;
-        let team = params.team.unwrap_or(plaid_stl::npm::NpmTeam::Developers);
+        let user = self.validate_npm_username(&params.user)?;
+        let team = params.team.unwrap_or("developers".to_string());
```
obelisk, line 495: This is a pretty strong default. Are we sure that it's the right one (generally speaking, for any org using this)?

obelisk: If you're confident, just leave a comment explaining why this is a default and you can resolve this comment.

obelisk (changes requested):

plaid-stl/src/npm/shared_structs/mod.rs
michelemin: That's a doc comment that applies to the module itself

plaid/src/apis/npm/npm_web_client.rs
michelemin: Right, I will add a comment pointing to https://docs.npmjs.com/about-developers-team

plaid/src/apis/npm/mod.rs
michelemin: Discussed offline and solved: for now, an error here is propagated to `Api::new`, we log an error, and Plaid's initialization continues

plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -0,0 +1,833 @@
...
+            .get(format!(
+                "{}/package/%40{}%2F{}/delete",
+                NPMJS_COM_URL, self.config.npm_scope, package
+            ))
+            .send()
+            .await
+            .map_err(|_| ApiError::NpmError(NpmError::FailedToDeletePackage))?;
+        // Safe unwrap: hardcoded data.
+        // TODO best to precompile regex? This is likely an infrequent operation.
+        let dsr_manifest_hash = Regex::new(r#"dsrManifestHash\"\s+?value=\"([a-z0-9]{64})\""#)
```
obelisk, line 343: This is recompiling the regex every time and should be a validator

plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -0,0 +1,833 @@
...
+            .client
+            .get(format!(
+                "{}/package/%40{}%2F{}/delete",
+                NPMJS_COM_URL, self.config.npm_scope, package
+            ))
+            .send()
+            .await
+            .map_err(|_| ApiError::NpmError(NpmError::FailedToDeletePackage))?;
+        // Safe unwrap: hardcoded data.
+        // TODO best to precompile regex? This is likely an infrequent operation.
```
obelisk, line 342: See comment below, would just put it with the rest of the validators.

plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -0,0 +1,833 @@
...
+
+        info!(
+            "Adding user [{}] to team [{}] on behalf of [{module}]",
+            user, team
+        );
+
+        let csrf_token = self
+            .get_csrftoken_from_cookies()
+            .map_err(|_| ApiError::NpmError(NpmError::WrongClientStatus))?;
+        let body = format!(r#"{{"csrftoken": "{}", "user": "{}"}}"#, &csrf_token, user);
```
obelisk, line 398: This should really be it's own structure then serialized properly. The validators *should* prevent any issues here, but only if they stay compatible with this kind of interpolation. Safer to just use `serde_json` serialize

plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -0,0 +1,833 @@
...
+            .await
+            .map_err(|_| ApiError::NpmError(NpmError::LoginFlowError))?;
+        let params: AddRemoveUserToFromTeamParams =
+            serde_json::from_str(params).map_err(|_| ApiError::BadRequest)?;
+        let user = self.validate_npm_username(&params.user)?;
+        let team = self.validate_npm_team_name(&params.team)?;
+
+        info!("Removing [{user}] from [{team}] on behalf of [{module}]");
+
+        let body = format!(
```
obelisk, line 424: Again, worried about this kind of interpolation.

plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -0,0 +1,833 @@
...
+    }
+
+    /// Remove a user from the npm organization
+    pub async fn remove_user_from_organization(
+        &self,
+        user: &str,
+        module: &str,
+    ) -> Result<i32, ApiError> {
+        #[derive(Serialize)]
+        struct RemoveUserPayload<'a> {
```
obelisk, line 449: Yes, I would pull this out so it can be used in other functions seeing as we have to do this a few times.

plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -0,0 +1,833 @@
...
+            .header("X-Spiferack", "1")
+            .send()
+            .await
+            .map_err(|_| ApiError::NpmError(NpmError::FailedToRetrievePackages))?
+            .json()
+            .await
+            .map_err(|_| ApiError::NpmError(NpmError::FailedToRetrievePackages))?;
+
+        // Get total number of packages that we will need to retrieve
+        let total_packages: u16 = serde_json::from_value(
```
obelisk, line 660: Interesting choice of `u16`. Is that a limit from NPM? If so where is that documented?

plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -0,0 +1,833 @@
...
+            .get_paginated_data_from_npm_website::<NpmPackageWithPermission>(
+                &format!(
+                    "{}/settings/{}/teams/team/{}/access",
+                    NPMJS_COM_URL, self.config.npm_scope, team
+                ),
+                total_packages,
+            )
+            .await?
+            .iter()
+            .filter(|v| v.permission.to_string() == params.permission.to_string())
```
obelisk, line 682: If `permission` is a struct you've defined you could implement the `Eq` trait and then check equality that way (you might also be able to derive it directly) and simplify this code.

obelisk (changes requested):

plaid/src/apis/npm/npm_web_client.rs
michelemin: Nice catch: changed to `u64`

plaid/src/apis/npm/npm_web_client.rs
michelemin: Done!

plaid/src/apis/npm/npm_web_client.rs
michelemin: Done: I made it a "special" validator because it's not sufficient to verify there is a match, but we also need to extract the token from the capturing group. Anyway, we are no longer recompiling the regex.

plaid/src/apis/npm/npm_web_client.rs
michelemin: Fixed, plus another one in the previous function

plaid/src/apis/npm/npm_web_client.rs
michelemin: Done. Turns out this struct is the same as the one we serialize for removing a user from a team. But I am reluctant to fuse them together because things might change in the future, and we don't want to couple these requests at all costs.

michelemin: That said, it's always about user removal from somewhere... I guess I will merge them 😄

plaid/src/apis/npm/npm_web_client.rs
michelemin: Fixed!

obelisk (approved):

PR #27, opened 2024-10-03
[npm] Support creating tokens for multiple packages at once

michelemin:
Instead of taking a single package name, take a vector of package names. They are all validated and used to define the scope of the token.

Also allow callers to specify the GH repo linked to the package, instead of assuming it is the same as package_name.

### DIFF
--- a/plaid/src/apis/npm/npm_web_client.rs
+++ b/plaid/src/apis/npm/npm_web_client.rs
@@ -8,7 +8,7 @@ use url::Url;
 use crate::apis::ApiError;
 
 use plaid_stl::npm::shared_structs::{
-    AddRemoveUserToFromTeamParams, CreateGranularTokenForPackageParams,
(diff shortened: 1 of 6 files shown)


plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -254,11 +261,7 @@ impl Npm {
                 .clone()
                 .map_or("Read and write".to_string(), |v| v.to_string()),
             selected_orgs: specs.selected_orgs.clone().unwrap_or(vec![]),
-            selected_packages: params
-                .specs
-                .selected_packages
-                .clone()
-                .unwrap_or(vec![scoped_package]),
+            selected_packages: scoped_packages,
```
michelemin, line 264: We take packages as an explicit parameter, so we must use those when defining the scope. It did not make sense that the token specs can contain an entirely different list of packages.

PR #28, opened 2024-10-04
Add npm APIs for retrieving details about / deleting / renewing a granular token

michelemin:
This PR adds
* An API to retrieve some details about an npm granular token (including the packages it can publish). Further processing is left to the calling rule.
* An API to delete npm granular tokens
* An API to renew an npm granular token "in place", i.e., overwrite it with a token that has the same name and same publish scope

### DIFF
--- a/plaid/src/apis/npm/npm_web_client.rs
+++ b/plaid/src/apis/npm/npm_web_client.rs
@@ -1,5 +1,5 @@
 use reqwest::cookie::CookieStore;
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::collections::HashMap;
 use totp_rs::{Algorithm, Secret, TOTP};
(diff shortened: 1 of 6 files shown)


plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -9,8 +9,10 @@ use crate::apis::ApiError;
 
 use plaid_stl::npm::shared_structs::{
     AddRemoveUserToFromTeamParams, CreateGranularTokenForPackagesParams,
-    InviteUserToOrganizationParams, ListPackagesWithTeamPermissionParams, NpmPackagePermission,
-    NpmToken, NpmUser, NpmUserRole, SetTeamPermissionOnPackageParams,
+    GetTokenPublishScopeParams, GranularTokenPackagesAndScopesPermission,
```
obelisk, line 12: At this point, probably just:

`use plaid_stl::npm::shared_structs::*;`

obelisk (approved):

plaid/src/apis/npm/npm_web_client.rs
```diff
@@ -761,6 +778,69 @@ impl Npm {
...
+                .map_err(|_| ApiError::NpmError(NpmError::FailedToGetTokenPublishScope))?
+                .get("tokenDetails")
+                .ok_or(ApiError::NpmError(NpmError::FailedToGetTokenPublishScope))?
+                .clone(),
+        )
+        .map_err(|_| ApiError::NpmError(NpmError::FailedToGetTokenPublishScope))?;
+
+        // Check a few things about the token before returning its publish scope
+
+        if token_details.expired
```
obelisk, line 822: This is counter intuitive to the caller, that just because a token is expired, the scopes disappear. I would wonder if we should build a more generic API to just return granular token details and work with specifics on the module side. Thoughts?

michelemin: Good idea and it would keep things easier on the runtime's side. So I will make a `get_token_details` API and return a `GranularTokenDetails` struct.
As for expired tokens, admittedly I don't know what happens when an NPM token expires. I have just created one which expires tomorrow so I'll wait and see.

obelisk (approved):

PR #29, opened 2024-10-04
Add mechanism to process modules non concurrently

wbssbw:
This PR adds a mechanism to mark rules that cannot be processed in parallel. In the configuration we can define single threaded rules like so:
```toml
[loading]
single_threaded_rules = ["testing_test.wasm"]
```

When one of these rules is seen in the execution loop, we check a list of "in flight" rules to see if it's already being processed. If it is, the message is sent back into the log queue for processing. If it's not, we add the rule name to the list of in flight rules and continue with execution. When execution is finished, we remove the rule from the in flight list.

### DIFF
--- a/plaid/src/executor/mod.rs
+++ b/plaid/src/executor/mod.rs
@@ -412,6 +412,11 @@ fn execution_loop(
                 Err(e) => Some(determine_error(e, computation_limit, &instance, &mut store)),
             };
 
+            // Mark the rule as complete
+            if let Some((_, is_locked)) = &plaid_module.concurrency_unsafe {
+                is_locked.store(false, Ordering::SeqCst);
+            }
+
             // If there was an error then log that it happened to the els
             if let Some(error) = error {
                 els.log_module_error(
@@ -426,11 +431,6 @@ fn execution_loop(
 
             // Update the persistent response
             update_persistent_response(&plaid_module, &env, &mut store)?;
-
-            // Mark the rule as complete
-            if let Some((_, is_locked)) = &plaid_module.concurrency_unsafe {
-                is_locked.store(false, Ordering::SeqCst);
-            }
         }
     }
     Err(ExecutorError::IncomingLogError)
--- a/plaid/src/executor/mod.rs
+++ b/plaid/src/executor/mod.rs
@@ -16,7 +16,6 @@ use wasmer_middlewares::metering::{get_remaining_points, MeteringPoints};
 
 use lru::LruCache;
 use std::collections::HashMap;
-use std::sync::atomic::Ordering;
 use std::sync::{Arc, RwLock};
 use std::thread::{self, JoinHandle};
 
@@ -344,21 +343,22 @@ fn execution_loop(
             // Mark this rule as currently being processed by locking the mutex
             // This lock will be dropped at the end of the iteration so we don't
             // need to handle unlocking it
-            if let Some((rule_in_use, is_locked)) = &plaid_module.concurrency_unsafe {
-                match rule_in_use.lock() {
-                    Ok(_) => is_locked.store(true, Ordering::SeqCst),
-                    Err(e) => {
+            let _lock = match plaid_module.concurrency_unsafe {
+                Some(ref mutex) => match mutex.lock() {
+                    Ok(guard) => Some(guard),
+                    Err(poisoned) => {
                         error!(
-                            "Failed to acquire lock on [{}] due to poisoning. This log will be discarded. Error: {e}.",
-                            plaid_module.name
-                        );
+                                "Failed to acquire lock on [{}] due to poisoning. This log will be discarded. Error: {}.",
+                                plaid_module.name, poisoned
+                            );
 
                         debug!("Clearing poison from the lock on [{}]", plaid_module.name);
-                        rule_in_use.clear_poison();
+                        mutex.clear_poison();
 
                         continue;
                     }
-                }
+                },
+                None => None,
             };
 
             // TODO @obelisk: This will quietly swallow locking errors on the persistent response
@@ -411,11 +411,6 @@ fn execution_loop(
                 }
                 Err(e) => Some(determine_error(e, computation_limit, &instance, &mut store)),
             };
-
-            // Mark the rule as complete
-            if let Some((_, is_locked)) = &plaid_module.concurrency_unsafe {
-                is_locked.store(false, Ordering::SeqCst);
-            }
 
             // If there was an error then log that it happened to the els
             if let Some(error) = error {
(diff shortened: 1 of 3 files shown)


plaid/src/executor/mod.rs
```diff
@@ -316,8 +318,58 @@ fn execution_loop(
             Some(module) => module,
         };
 
+        // If ANY of the modules that operate on this log type cannot be executed in parallel, we need to check
```
obelisk, line 321: This is not something I had considered: having a single threaded rule in a message type makes all things processing that log single threaded.

That is going to result in larger than expected performance hits when a non-concurrent rule is added to a logging channel.

plaid/src/executor/mod.rs
```diff
@@ -316,8 +318,58 @@ fn execution_loop(
...
+            };
+
+            // If this rule is currently being processed, send the message back into the execution queue
+            // This allows us to guarantee that the rule will only be executed on 1 thread at any given time
+            if execution_modules
+                .iter()
+                .any(|module| strif.contains(&module.name))
+            {
+                if let Err(e) = sender.send(message.clone()) {
+                    error!("Failed to send message back into execution queue. Error: {e}");
```
obelisk, line 350: This should mention that the log is now lost as well

plaid/src/executor/mod.rs
```diff
@@ -316,8 +318,58 @@ fn execution_loop(
...
+            .iter()
+            .any(|module| !module.parallel_execution_enabled)
+        {
+            let strif = match single_threaded_rules_in_flight.read() {
+                Ok(strif) => strif,
+                Err(e) => {
+                    // .read() will return an error if the lock is poisoned (which should never happen). If this occurs,
+                    // we'll clear the poison and send the message back into the execution queue
+
+                    error!("Failed to acquire read lock. Error: {e}");
```
obelisk, line 333: This should mention that the mutex was poisoned. Just for clarity.

plaid/src/executor/mod.rs
```diff
@@ -316,8 +318,58 @@ fn execution_loop(
...
+                    single_threaded_rules_in_flight.clear_poison();
+
+                    if let Err(e) = sender.send(message.clone()) {
+                        error!("Failed to send message back into execution queue. Error: {e}");
```
obelisk, line 337: We should mention the log was lost.

plaid/src/executor/mod.rs
obelisk: There should also be an info or trace log here that a log was delayed due to parallel execution prevention.

plaid/src/executor/mod.rs
```diff
@@ -316,8 +318,58 @@ fn execution_loop(
...
         for plaid_module in execution_modules {
+            if !plaid_module.parallel_execution_enabled {
+                match single_threaded_rules_in_flight.write() {
+                    Ok(mut strif) => {
+                        strif.insert(plaid_module.name.clone());
+                    }
+                    // Poisons should never happen because: An RwLock is poisoned whenever a writer panics while holding an exclusive lock.
+                    // We hold the write lock in two locations, both with proper error handling to ensure that it will never panic.
+                    Err(e) => {
+                        error!("Failed to acquire write lock. Error: {e}");
```
obelisk, line 367: Log lost

plaid/src/executor/mod.rs
```diff
@@ -369,6 +421,18 @@ fn execution_loop(
                 Err(e) => Some(determine_error(e, computation_limit, &instance, &mut store)),
             };
 
+            // The rule has finished executing. Remove it from the list of currently running rules
+            if !plaid_module.parallel_execution_enabled {
+                match single_threaded_rules_in_flight.write() {
+                    Ok(mut strif) => {
+                        strif.insert(plaid_module.name.clone());
```
obelisk, line 428: 95oldj

plaid/src/loader/mod.rs
```diff
@@ -98,6 +102,7 @@ pub struct PlaidModule {
     pub cache: Option<Arc<RwLock<LruCache<String, String>>>>,
     /// See the PersistentResponse type.
     pub persistent_response: Option<PersistentResponse>,
+    pub parallel_execution_enabled: bool,
```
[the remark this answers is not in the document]
obelisk: I think this should just be an `Option<Mutex<()>>`.

If it is `None`, then this supports parallel execution. If it is `Some(Mutex<()>)` then we can try to lock it, if we can't then it's in use and we throw it to the back. If we can, we own it exclusively and can execute the module.

obelisk (changes requested):

obelisk:
PR description is outdated

plaid/src/loader/mod.rs
```diff
@@ -108,6 +118,15 @@ impl PlaidModule {
             .flatten()
     }
 
+    /// Method to check if the module is currently locked
+    pub fn is_locked(&self) -> bool {
+        if let Some(mutex) = &self.concurrency_unsafe {
+            // Return true if try_lock fails (meaning the mutex is locked), otherwise false
+            return mutex.try_lock().is_err();
```
obelisk, line 125: I think there are other errors here than just `WouldBlock`. If the Mutex is poisoned, I don't think we'd ever detect that because this function would always appear to return "Locked" and thus we'd never hit the poison resolution case above.

obelisk: I think the solution here is to log the poison and clear it here.

plaid/src/executor/mod.rs
```diff
@@ -316,8 +317,50 @@ fn execution_loop(
...
+            // This allows us to guarantee that the rule will only be executed on 1 thread at any given time
+
+            info!(
+                "Message of type [{}] is currently being processed by another rule. Requeueing the message for later execution.",
+                message.type_
+            );
+
+            if let Err(e) = sender.send(message.clone()) {
+                error!(
+                    "Failed to requeue message of type [{}]. It is permanently lost. Error: {e}",
```
obelisk, line 333: Do we want to print the actual log here?

obelisk: Seeing as it's been lost

plaid/src/executor/mod.rs
```diff
@@ -316,8 +317,50 @@ fn execution_loop(
...
+            continue;
+        };
+
         // For every module that operates on that log type
         for plaid_module in execution_modules {
+            // Mark this rule as currently being processed by locking the mutex
+            // This lock will be dropped at the end of the iteration so we don't
+            // need to handle unlocking it
+            let _lock = match plaid_module.concurrency_unsafe {
+                Some(ref mutex) => match mutex.lock() {
```
obelisk, line 347: I believe this can deadlock if Plaid ever decided to execute rules in random orders (maybe for performance reasons, or because in the future the rules structure can be updated at runtime).

That could allow some mutexes to be acquired, while waiting for others. Happening between two threads, they could lock and prevent log processing.

obelisk (changes requested):

obelisk (approved):

PR #30, opened 2024-10-07
Support searching for file in GitHub

michelemin:
Add an API that allows one to search for files with a given name in GitHub. The API also supports specifying selection criteria that determine which results should be kept or discarded.

### DIFF
--- a/plaid-stl/src/github/mod.rs
+++ b/plaid-stl/src/github/mod.rs
@@ -104,7 +104,9 @@ impl FileSearchResultItem {
 #[derive(Serialize, Deserialize)]
 /// Specifies criteria according to which results returned
 /// by the API should be included or discarded.
-pub struct FileSearchSelectionCriteria {
+pub struct CodeSearchCriteria {
+    /// Keep only results from this GH org
+    pub only_from_org: Option<String>,
     /// Keep only results from these repositories (mutually exclusive with `discard_from_repos`)
     pub only_from_repos: Option<Vec<String>>,
     /// Discard files from these repositories (mutually exclusive with `only_from_repos`)
@@ -951,39 +953,62 @@ pub fn configure_secret(
 /// - `filename`: The name of the files to search, e.g., "README.md"
 /// - `org`: The name of the GitHub organization
 /// - `selection_criteria`: An optional `FileSearchSelectionCriteria` object with additional criteria
-pub fn search_file_in_org_code(
+pub fn search_file(
     filename: impl Display,
-    org: impl Display,
-    repo: Option<impl Display>,
-    selection_criteria: Option<&FileSearchSelectionCriteria>,
+    search_criteria: Option<&CodeSearchCriteria>,
 ) -> Result<Vec<FileSearchResultItem>, PlaidFunctionError> {
     extern "C" {
-        new_host_function_with_error_buffer!(github, search_file_in_org_code);
+        new_host_function_with_error_buffer!(github, search_file);
     }
+
     let mut params: HashMap<&str, String> = HashMap::new();
     params.insert("filename", filename.to_string());
-    params.insert("org", org.to_string());
-    if let Some(repo) = repo {
-        params.insert("repo", repo.to_string());
+
+    // If we are given selection criteria, then we divide them between
+    //
+    // * Those that can be baked directly into the GitHub search query, thus making the overall search more
+    // efficient (because less results are returned). These are passed to the API.
+    //
+    // * Those that have to be (or are better) evaluated module-side. These are not passed to the API and
+    // are processed later here.
+
+    if let Some(criteria) = search_criteria {
+        // Check for mutually exclusive criteria
+        if criteria.only_from_repos.is_some() && criteria.discard_from_repos.is_some() {
+            return Err(PlaidFunctionError::ErrorCouldNotSerialize);
+        }
+
+        if let Some(org) = &criteria.only_from_org {
+            // Search only inside an organization
+            params.insert("org", org.clone());
+
+            if let Some(repos) = &criteria.only_from_repos {
+                if repos.len() == 1 {
+                    // Special case: search only in a repository
+                    params.insert("repo", repos[0].clone());
+                }
+            }
+        }
     }
 
     let mut search_results = Vec::<FileSearchResultItem>::new();
     let mut page = 0;
+
+    // Use a larger page size to make less requests and reduce chances of hitting the rate limit
+    params.insert("per_page", "100".to_owned());
 
     const RETURN_BUFFER_SIZE: usize = 1 * 1024 * 1024; // 1 MiB
 
     loop {
         page += 1;
         params.insert("page", page.to_string());
-        // use a larger page size to make less requests and avoid hitting the rate limit
-        params.insert("per_page", "100".to_owned());
 
         let request = serde_json::to_string(&params).unwrap(); // safe unwrap
 
         let mut return_buffer = vec![0; RETURN_BUFFER_SIZE];
 
         let res = unsafe {
-            github_search_file_in_org_code(
+            github_search_file(
                 request.as_bytes().as_ptr(),
                 request.as_bytes().len(),
                 return_buffer.as_mut_ptr(),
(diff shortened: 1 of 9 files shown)


obelisk:
Descriptions please.

obelisk:
🗞️

plaid-stl/src/github/mod.rs
```diff
@@ -2,7 +2,7 @@ use std::{collections::HashMap, fmt::Display};
 
 use serde::{Deserialize, Serialize};
 
-use crate::PlaidFunctionError;
+use crate::{plaid::print_debug_string, PlaidFunctionError};
```
obelisk, line 5: We shouldn't be importing this here.

plaid-stl/src/github/mod.rs
```diff
@@ -29,6 +29,101 @@ pub struct RepositoryCollaborator {
...
+            &self.repository.owner.login,
+            &self.repository.name,
+            &self.path,
+            reference,
+        )
+        .map_err(|_| PlaidFunctionError::InternalApiError)?;
+        let content = serde_json::from_str::<GithubFileContent>(&content)
+            .map_err(|e| {
+                print_debug_string(&format!("ERROR1 {e}"));
+                panic!()
```
obelisk, line 93: We shouldn't panic here

plaid-stl/src/github/mod.rs
```diff
@@ -843,12 +938,152 @@ pub fn configure_secret(
...
+
+/// Search for files with given filename in a GitHub organization's code.
+/// If additional selection criteria are given, these are used to decide whether
+/// results are selected or discarded.
+///
+/// **Arguments:**
+/// - `filename`: The name of the files to search, e.g., "README.md"
+/// - `org`: The name of the GitHub organization
+/// - `selection_criteria`: An optional `FileSearchSelectionCriteria` object with additional criteria
+pub fn search_file_in_org_code(
```
obelisk, line 958: Do we want to make this more general and just allow general GitHub search? Even perhaps outside of the org by wrapping the GitHub search API?

plaid/src/apis/github/code.rs
```diff
@@ -0,0 +1,61 @@
...
+            // GitHub supports up to 100 results per page
+            return Err(ApiError::BadRequest);
+        }
+
+        info!(
+            "Finding all files called [{filename}] in organization [{org}] on behalf of [{module}]"
+        );
+
+        // Build the search query
+        let query = urlencoding::encode(&format!("{} in:path org:{}", filename, org)).to_string();
```
obelisk, line 41: I'm thinking we should provide a generic search API. Any thoughts on the pros and cons?

obelisk (changes requested):

plaid/src/apis/github/code.rs
michelemin: Yes, that's a fair point. I was considering this but then I stumbled on the rate limitation for the search code endpoint (see [here](https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#rate-limit)).
We get only 10 requests per minute, so I increased the page size to its max, in an effort to make as few calls as we can.

Basically, my fear is that if we allow very generic searches, rules might start experiencing failures and everything kind of grinds to a halt when it comes to GH searches. If we narrow down the scope, we have more of a chance. 

What do you think?

plaid-stl/src/github/mod.rs
michelemin: See reply to other comment about search endpoint rate limitation.

plaid-stl/src/github/mod.rs
michelemin: Yep, leftover debugging code 😞

plaid/src/apis/github/code.rs
michelemin: @obelisk I added support for searching only in a repository, instead of searching in the entire org and _then_ filtering. So we make less calls and we should stay farther away from rate limitations.

plaid-stl/src/github/mod.rs
```diff
@@ -29,6 +29,99 @@ pub struct RepositoryCollaborator {
...
+    pub content: String,
+    pub encoding: String,
+}
+
+// END OF Structs used for deserializing GH's response when searching for code
+
+impl FileSearchResultItem {
+    /// Retrieve the content of the search result
+    pub fn retrieve_raw_content(&self) -> Result<String, PlaidFunctionError> {
+        let reference_regex = regex::Regex::new(r"^.*?ref=([a-f0-9]{40})$").unwrap(); // TODO improve
```
obelisk, line 76: This is going to make this call very expensive because compilation of regexes is intensive. If there's no way around it then it's fine (regex doesn't support static regexes yet) but it might be something we need to watch for when running simple rules using it because they could run out of compute :p

plaid-stl/src/github/mod.rs
```diff
@@ -29,6 +29,99 @@ pub struct RepositoryCollaborator {
...
+#[derive(Serialize, Deserialize)]
+/// Specifies criteria according to which results returned
+/// by the API should be included or discarded.
+pub struct CodeSearchCriteria {
+    /// Keep only results from this GH org
+    pub only_from_org: Option<String>,
+    /// Keep only results from these repositories (mutually exclusive with `discard_from_repos`)
+    pub only_from_repos: Option<Vec<String>>,
+    /// Discard files from these repositories (mutually exclusive with `only_from_repos`)
+    pub discard_from_repos: Option<Vec<String>>,
```
obelisk, line 113: Then this should be an enum so only one can be passed

plaid-stl/src/github/mod.rs
```diff
@@ -843,12 +936,173 @@ pub fn configure_secret(
...
+    let res =
+        unsafe { github_configure_secret(request.as_bytes().as_ptr(), request.as_bytes().len()) };
 
     match res {
         0 => Ok(()),
         x => Err(x.into()),
     }
 }
+
+/// Search for files with given filename in a GitHub organization's code.
```
obelisk, line 948: It's more generic than this now

plaid-stl/src/github/mod.rs
```diff
@@ -843,12 +936,173 @@ pub fn configure_secret(
...
+/// If additional selection criteria are given, these are used to decide whether
+/// results are selected or discarded.
+///
+/// **Arguments:**
+/// - `filename`: The name of the files to search, e.g., "README.md"
+/// - `org`: The name of the GitHub organization
+/// - `selection_criteria`: An optional `FileSearchSelectionCriteria` object with additional criteria
+pub fn search_file(
```
obelisk, line 956: Nit. 
search_file => implies searching within a file
search_for_file => implies looking for a file

plaid-stl/src/github/mod.rs
```diff
@@ -843,12 +936,173 @@ pub fn configure_secret(
...
+    // If we are given selection criteria, then we divide them between
+    //
+    // * Those that can be baked directly into the GitHub search query, thus making the overall search more
+    // efficient (because less results are returned). These are passed to the API.
+    //
+    // * Those that have to be (or are better) evaluated module-side. These are not passed to the API and
+    // are processed later here.
+
+    if let Some(criteria) = search_criteria {
+        // Check for mutually exclusive criteria
```
obelisk, line 976: Making this an enum would solve this issue

plaid/src/apis/github/code.rs
```diff
@@ -0,0 +1,85 @@
...
+                    ),
+                    Some(repo) => info!(
+                        "Finding all files called [{filename}] in repository [{org}/{repo}] on behalf of [{module}]"
+                    ),
+                }
+        }
+
+        // Build the search query
+        let query = match org {
+            None => format!("{} in:path", filename),
```
obelisk, line 59:
```suggestion
            None => format!("{filename} in:path"),
```

plaid/src/apis/github/code.rs
```diff
@@ -0,0 +1,85 @@
...
+            Some(org) => match repo {
+                None => format!("{} in:path org:{}", filename, org),
```
obelisk, line 61:
```suggestion
                None => format!("{filename} in:path org:{org}"),
```

plaid/src/apis/github/code.rs
```diff
@@ -0,0 +1,85 @@
...
+                Some(repo) => format!("{} in:path repo:{}/{}", filename, org, repo),
```
obelisk, line 62:
```suggestion
                Some(repo) => format!("{filename} in:path repo:{org}/{repo}"),
```

obelisk (changes requested):
I would split the module side filtering out into its own function for cleanliness. I think it's still fine to call it from within this function, but I can see this being useful outside of just this function. For example, maybe you want to make a query, see how many results you get back, before filtering.

Breaking it out into its own function also allows you to run multiple filters on the raw data without querying more than once.

plaid-stl/src/github/mod.rs
michelemin: Done

PR #36, opened 2024-10-22
Implement benchmarking system to measure and compare rule execution data

wbssbw:
Currently, we're unable to quantify how subtle changes to a rule or the Plaid runtime affect performance. This PR implements a mechanism to benchmark and compare Plaid rules, enabling developers and plaidmins to maximize their instance's performance. When enabled, it logs rule execution metadata (rule name, execution time, computation used) to a task that aggregates this data based on rule name. On shutdown, the benchmarking system writes this data out to a file.

Triggering shutdown of the server is done with `CTRL+C` or by sending a `SIGTERM` signal. Plaid is configured to listen for this signal and begin a graceful shutdown. Right now, graceful shutdown means waiting for the `benchmark_loop` task to exit. This allows us to guarantee that our benchmarking data will be written to a file prior to exiting. However, graceful shutdown allows us to do more in the future. We could extend this idea to the execution loop and only shutdown once the message queue is empty, therefore guaranteeing that all logs will be processed.

### Output
The benchmarking system outputs a file like:
```
Performance Report:
Module: testing_test_release_default.wasm
	Runs: 3510
	Average Computation Used: 4752.00
	Average Execution Time (microseconds): 2012

Module: testing_test.wasm
	Runs: 3510
	Average Computation Used: 31317.00
	Average Execution Time (microseconds): 1967
```

In this simple example, we can see that when we compile the `testing_test.wasm` rule with the `--release` flag, we reduce the computation used by about 85%! That's a big difference.

### DIFF
--- a/plaid/src/benchmark/mod.rs
+++ b/plaid/src/benchmark/mod.rs
@@ -1,0 +1,168 @@
+use std::collections::HashMap;
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
+use std::time::Duration;
+use tokio::fs::File;
+use tokio::io::AsyncWriteExt;
+
+use crossbeam_channel::Receiver;
+use serde::Deserialize;
+
+#[derive(Deserialize)]
+pub struct Benchmarking {
+    /// The full path to the output file where benchmarking metrics should be written
+    #[serde(default = "default_results_file_path")]
+    output_file_path: String,
+}
+
+/// Default file path for benchmarking results if none is provided in the config
+fn default_results_file_path() -> String {
+    format!(
+        "{}/benchmark_results/metrics.txt",
+        env!("CARGO_MANIFEST_DIR")
+    )
+}
+
+/// Metadata about a rule's execution
+pub struct ModulePerformanceMetadata {
+    /// The name of the module
+    pub module: String,
+    /// Time (in microseconds) for execution to complete
+    pub execution_time: u128,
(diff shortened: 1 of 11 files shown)


plaid/src/benchmark/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+impl Benchmarking {
+    pub async fn benchmark_loop(
+        &self,
+        receiver: Receiver<ModulePerformanceMetadata>,
+        shutdown_initiated: Arc<AtomicBool>,
+    ) {
+        let mut aggregate_performance_metadata = HashMap::new();
+
+        // Benchmarking loop runs until shutdown_initiated is set to true
+        while !shutdown_initiated.load(Ordering::SeqCst) {
```
obelisk, line 95: Is this needed? It seems to me that when a graceful shutdown occurs the sender for the below will be dropped causing looping to cease here anyway.

You also have a task here which has no awaits in it. This is basically going to pin an entire tokio thread. Should we not just await on the `recv` and let it shutdown by dropping the sender?

wbssbw: Yeah I think it's probably possible to drop that channel and use it to trigger report generation.

And yeah, this pretty much pins the thread in its current state - will see if there's a way to make it less resource intensive

wbssbw:
Closing in favor of https://github.com/obelisk/plaid/pull/57

PR #37, opened 2024-10-23
Support adding error context for Plaid runtime + panic hook

michelemin:
This PR adds the following:
* A way for modules to give the Plaid runtime more context about errors they encounter. This is achieved by setting an error context (a `String`) which is then included in the error returned by the runtime.
* A simple panic hook (see [docs](https://doc.rust-lang.org/std/panic/fn.set_hook.html)) that, when a panic occurs, sets an appropriate context.

Note - The panic hook has been tested and works for things like explicit `panic!` and `index out of bounds`, but does not seem to work for OOM errors. For this reason, the explanation for an unknown error defaults to "This is probably an OOM error", since this is the only case we know of where the hook is not triggered and an explanation is not available.

### Impact
**! This change forces all rules to be modified and recompiled !** in order to support the new error management mechanism.

The return type of the `main` function must be `Result<(), XYZ>` where `XYZ` is something (most likely an enum) that implements `Error` (empty impl is fine) and `Display`.

### DIFF
--- a/plaid-stl/src/lib.rs
+++ b/plaid-stl/src/lib.rs
@@ -79,56 +79,30 @@ pub mod messages;
 //pub use ::plaid::LogSource;
 
 #[macro_export]
-macro_rules! setup_non_allocating_panic_mgmt {
-    () => {
-        use std::fmt::Write;
-
-        // A simple stack-allocated buffer with fixed size
-        struct NoAllocBuffer<'a> {
-            buf: &'a mut [u8],
-            pos: usize,
-        }
-
-        impl<'a> NoAllocBuffer<'a> {
-            fn new(buf: &'a mut [u8]) -> Self {
-                NoAllocBuffer { buf, pos: 0 }
-            }
-        }
-
-        impl<'a> std::fmt::Write for NoAllocBuffer<'a> {
-            fn write_str(&mut self, s: &str) -> std::fmt::Result {
-                let bytes = s.as_bytes();
-                let remaining_space = self.buf.len().saturating_sub(self.pos);
-                let to_write = std::cmp::min(remaining_space, bytes.len());
-
-                if to_write > 0 {
-                    self.buf[self.pos..self.pos + to_write].copy_from_slice(&bytes[..to_write]);
-                    self.pos += to_write;
-                }
-
-                if to_write < bytes.len() {
-                    return Err(std::fmt::Error); // Buffer overflow
-                }
-
-                Ok(())
-            }
-        }
-    };
-}
-
-#[macro_export]
 macro_rules! set_panic_hook {
     () => {
-        std::panic::set_hook(Box::new(|panic_info| {
-            // plaid::set_error_context(&panic_info.to_string());
-            let mut buffer = [0u8; 512]; // Stack-allocated buffer
-            let mut writer = NoAllocBuffer::new(&mut buffer);
+        use std::sync::{Arc, Mutex};
+        let buffer = Arc::new(Mutex::new([0u8; 512]));
+        let buffer_clone = Arc::clone(&buffer);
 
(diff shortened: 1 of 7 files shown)


plaid-stl/src/lib.rs
```diff
@@ -111,10 +111,17 @@ macro_rules! entrypoint {
                 Ok(s) => s,
                 Err(_) => return -2,
             };
+            
+            std::panic::set_hook(Box::new(|panic_info| {
+                plaid::set_error_context(panic_info.to_string());
```
obelisk, line 116: This is always going to fail if you receive an OOM since it allocates.

plaid-stl/src/plaid/mod.rs
```diff
@@ -156,3 +158,14 @@ pub fn get_response() -> Result<String, PlaidFunctionError> {
         Err(_) => Err(PlaidFunctionError::ParametersNotUtf8),
     }
 }
+
+/// Give the runtime more context about an error encountered during execution
+pub fn set_error_context(context: impl Display) {
```
obelisk, line 163: In this case, I would have this take an &str to prevent allocations when calling this function

plaid/src/executor/mod.rs
```diff
@@ -369,7 +375,12 @@ fn execution_loop(
             let error = match entrypoint.call(&mut store) {
                 Ok(n) => {
                     if n != 0 {
-                        Some(ModuleExecutionError::ModuleErrorCode(n))
+                        Some(ModuleExecutionError::ModuleError(
+                            env.as_ref(&store)
+                                .execution_error_context
+                                .clone()
+                                .unwrap_or("> empty <".to_string()),
```
obelisk, line 382: I'd just use None instead of `> empty <` unless you have a further reason to depart from the standard `Option` verbiage.

obelisk (changes requested):

plaid-stl/src/lib.rs
```diff
@@ -78,9 +78,68 @@ pub mod messages;
...
+            pos: usize,
+        }
+
+        impl<'a> NoAllocBuffer<'a> {
+            fn new(buf: &'a mut [u8]) -> Self {
+                NoAllocBuffer { buf, pos: 0 }
+            }
+        }
+
+        impl<'a> std::fmt::Write for NoAllocBuffer<'a> {
```
michelemin, line 98: For clarity: I asked ChatGPT to come up with a non-allocating version of the code I had before.
I tested this code but we still end up with "This is probably an OOM", which is the default message when everything else fails.

plaid-stl/src/lib.rs
```diff
@@ -78,9 +78,68 @@ pub mod messages;
...
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! set_panic_hook {
+    () => {
+        std::panic::set_hook(Box::new(|panic_info| {
+            // plaid::set_error_context(&panic_info.to_string());
+            let mut buffer = [0u8; 512]; // Stack-allocated buffer
```
obelisk, line 124: I'm worried this could still be heap allocated depending on the compiler. I would move this allocation outside the closure then capture it. That will ensure that even if it is heap allocated, it happens before memory becomes a problem (at least in theory).

obelisk (changes requested):

plaid-stl/src/lib.rs
```diff
@@ -78,9 +78,41 @@ pub mod messages;
 
 //pub use ::plaid::LogSource;
 
+#[macro_export]
+macro_rules! set_panic_hook {
```
obelisk, line 82: I think this can be simplified but we can come back to it

obelisk (approved):

PR #40, opened 2024-10-25
Fix return value for npm token generation

michelemin:
The problem is that `Value`s are returned in JSON format. This was in turn causing problems downstream.
The temporary solution is for the caller downstream to remove all `"` characters from the result, but this is the permanent fix.


obelisk (approved):
I'm assuming this works because before `to_string()` was causing the value to be serialized where as this is requesting the `str` value of the `Value` which will fail when the type is not a string.

It might be cleaner in the future to deserialize into a structure, or more clear to check `is_string` first but this works and there isn't a pressing reason to change it.

PR #41, opened 2024-10-28
Add log_back_unlimited API

michelemin:
This PR adds an API for rules to log back with unlimited budget.

Until now, rules could log back either with 0 budget for further invocations, or with a fixed budget. This would not allow one to implement a use case where something needs to continue forever at regular intervals.

Instead, with this change, a rule that picks up a message with unlimited budget can trigger further invocations with unlimited budget.

### Backward compatibility
The change is backward compatible: existing APIs are left untouched and a new API is added.

### Security considerations
Malicious rules could use this feature to try and mount a DOS attack against the runtime, by triggering an endless sequence of log backs. In the future, the problem might be addressed by allowing one to specify which rules are trusted to use this feature.

### Extra notes
There is no API for a rule to query the number of invocations left. If there was, one could approximate the use case above by starting with a high budget and, each time, logging back with budget `what_we_had_before - 1`. Such an API might be added in the future.

### DIFF
--- a/plaid-stl/src/plaid/mod.rs
+++ b/plaid-stl/src/plaid/mod.rs
@@ -23,6 +23,35 @@ pub fn log_back(type_: &str, log: &[u8], delay: u32) -> Result<(), i32> {
     log_back_with_budget(type_, log, delay, 0)
 }
 
+/// Send a log to the logback system with unlimited extra budget to trigger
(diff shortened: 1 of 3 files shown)


obelisk (approved):
It's a bit annoying to reuse the `LogbacksAllowed` type when it's actually log backs requested but I don't have a better solution.

Issue #6, opened 2024-07-11
GitHub Data Generator Should Be App Compatible


obelisk:
Having the data generator as it is right now means that you need to cut an FPAT (or CPAT) in addition to the GitHub app. This is needless and leads to system downtime when the FPAT expires.

The data generator and GitHub API should be able to share a GitHub App configuration

PR #48, opened 2024-11-04
Modify GitHub data generator to work with GitHub app credentials

wbssbw:
This makes the GitHub data generator compatible with GitHub app creds, removing the need for FPATs as described in [this issue](https://github.com/obelisk/plaid/issues/6)

It could be nice to have the API and data generator share the same GitHub client, but that would require breaking changes to the config to move the authentication settings outside of the API and data generators section.

### DIFF
--- a/plaid/src/data/github/mod.rs
+++ b/plaid/src/data/github/mod.rs
@@ -101,7 +101,7 @@ where
         "git" => Ok(LogType::Git),
         "all" => Ok(LogType::All),
         _ => Err(serde::de::Error::custom(
-            "Invalid log tpye provided. Expected one of All, Git, Web",
+            "Invalid log type provided. Expected one of All, Git, Web",
         )),
     }
 }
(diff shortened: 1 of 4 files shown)


obelisk (commented):
This looks good. Can you update the example plaid.toml so I can easily see how this is configured?

plaid/src/data/github/mod.rs
```diff
@@ -55,97 +63,140 @@ impl Ord for GithubAuditLog {
...
+    D: serde::Deserializer<'de>,
+{
+    let log_type = String::deserialize(deserializer)?.to_lowercase();
+
+    match log_type.as_str() {
+        "web" => Ok(LogType::Web),
+        "git" => Ok(LogType::Git),
+        "all" => Ok(LogType::All),
+        _ => Err(serde::de::Error::custom(
+            "Invalid log tpye provided. Expected one of All, Git, Web",
```
obelisk, line 104: Nit typo

obelisk (approved):

obelisk:
We are going to need to test this heavily but otherwise looks good

plaid/src/data/github/mod.rs
wbssbw: Fixed!

obelisk (approved):

PR #51, opened 2024-11-06
Check status code for each call to npm website

michelemin:
Whenever we make a call to the npm website, we check that the status code is the expected one (usually 200, but in one case it returns 302).


obelisk (approved):
Looks good, and returning complex errors to rules so they can make sensible decisions is definitely the future.

PR #52, opened 2024-11-06
Parse more fields for Copilot list seats API

timweri:
Changes in order of importance:
1. I need more parsed fields in the List Copilot Seats API to tell when seats are team-based and when seats have already been queued for removal.
2. Update datetime deserializers
3. Remove an unused line flagged by the compiler.


plaid-stl/src/npm/shared_structs/mod.rs
```diff
@@ -169,7 +168,7 @@ pub struct NpmToken {
     pub token_name: Option<String>,
     pub token_type: Option<String>,
     // Deserialize the date field from the ISO 8601 format
-    #[serde(default, deserialize_with = "deserialize_option_timestamp")]
+    #[serde(with = "chrono::serde::ts_seconds_option")]
```
timweri, line 171: @michelemin  `chrono` has a built in function for this. Do you bless this or should I leave your portion as is?

plaid-stl/src/github/mod.rs
```diff
@@ -40,20 +40,31 @@ pub struct CopilotAssignee {
...
-    pub created_at: Option<DateTime<Utc>>,
-    // Deserialize the date field from the ISO 8601 format
-    #[serde(default, deserialize_with = "deserialize_option_timestamp")]
-    pub updated_at: Option<DateTime<Utc>>,
+    // Deserialize the datetime field from the ISO 8601 format
+    #[serde(with = "chrono::serde::ts_seconds")]
+    pub created_at: DateTime<Utc>,
+    // Deserialize the datetime field from the ISO 8601 format
+    #[serde(with = "chrono::serde::ts_seconds")]
+    pub updated_at: DateTime<Utc>,
```
timweri, line 67: According to the API schema, `created_at` and `updated_at` are not null.

plaid-stl/src/datetime.rs
```diff
@@ -1,18 +1,64 @@
...
+    // throwing parsing errors
+    if s.is_empty() {
+        return Ok(None)
+    }
+
+    Ok(Some(
+        DateTime::parse_from_rfc3339(&s)
+            .map_err(serde::de::Error::custom)?
+            .with_timezone(&Utc)
+    ))
```
timweri, line 40: @michelemin  If the string is empty, I return None. Before, if the string is not empty and parsing fails, the deserializer returns None, which I think is misleading. Now, if parsing fails on an non-empty string, the deserializer will throw an error.

michelemin (approved):
LGTM

plaid-stl/src/datetime.rs
```diff
@@ -1,18 +1,64 @@
-use chrono::{DateTime, Utc};
+use chrono::{NaiveDate, Utc, DateTime};
 use serde::{Deserialize, Deserializer};
 
-// Custom deserializer for an optional DateTime<Utc>
-pub fn deserialize_option_timestamp<'de, D>(deserializer: D) -> Result<Option<DateTime<Utc>>, D::Error>
+// Custom deserializer from a ISO 8601 string to a DateTime<Utc>
+pub fn deserialize_rfc3339_timestamp<'de, D>(deserializer: D) -> Result<DateTime<Utc>, D::Error>
```
obelisk, line 5: Interesting choice in the comment to refer to it as ISO 8601 and the function RFC3339?

obelisk (approved):

plaid-stl/src/datetime.rs
timweri: :P I want to call it ISO 8601 but the chrono function uses RFC 3339

PR #57, opened 2024-11-22
Implement benchmarking system

wbssbw:
Currently, we're unable to quantify how subtle changes to a rule or the Plaid runtime affect performance. This PR implements a mechanism to benchmark and compare Plaid rules, enabling developers and plaidmins to maximize their instance's performance. When enabled, it logs rule execution metadata (rule name, execution time, computation used) to a task that aggregates this data based on rule name. On shutdown, the benchmarking system writes this data out to a file.

Triggering shutdown of the server is done with `CTRL+C` or by sending a `SIGTERM` signal. Plaid is configured to listen for this signal and begin a graceful shutdown. Right now, graceful shutdown means waiting for the `benchmark_loop` task to exit. This allows us to guarantee that our benchmarking data will be written to a file prior to exiting. However, graceful shutdown allows us to do more in the future. We could extend this idea to the execution loop and only shutdown once the message queue is empty, therefore guaranteeing that all logs will be processed.

### Output
The benchmarking system outputs a file like:
```
Performance Report:
Module: testing_test_release_default.wasm
	Runs: 3510
	Average Computation Used: 4752.00
	Average Execution Time (microseconds): 2012

Module: testing_test.wasm
	Runs: 3510
	Average Computation Used: 31317.00
	Average Execution Time (microseconds): 1967
```

In this simple example, we can see that when we compile the `testing_test.wasm` rule with the `--release` flag, we reduce the computation used by about 85%! That's a big difference.

### DIFF
--- a/runtime/plaid/Cargo.toml
+++ b/runtime/plaid/Cargo.toml
@@ -56,6 +56,7 @@ tokio-tungstenite = { version = "0.23.1", features = [
 futures-util = "0.3.30"
 aws-sdk-kms = { version = "1.41.0", optional = true }
 aws-config = { version = "1.5.5", optional = true }
+tokio-util = "0.7.12"
 
 [[example]]
 name = "github-tailer"
--- a/runtime/plaid/Cargo.toml
+++ b/runtime/plaid/Cargo.toml
@@ -42,6 +42,10 @@ sled = "0.34.7"
 tar = "0.4.41"
 time = "0.3"
 tokio = { version = "1", features = ["full"] }
+tokio-util = "0.7.12"
+tokio-tungstenite = { version = "0.23.1", features = [
+    "rustls-tls-native-roots",
+] }
 toml = "0.5"
 totp-rs = "5.6.0"
 url = "2.5.2"
@@ -50,13 +54,9 @@ warp = { version = "0.3", features = ["tls"] }
 wasmer = { version = "4", default-features = false, features = ["cranelift"] }
 wasmer-middlewares = "4"
 jsonwebtoken = { version = "9.2" }
-tokio-tungstenite = { version = "0.23.1", features = [
-    "rustls-tls-native-roots",
-] }
 futures-util = "0.3.30"
 aws-sdk-kms = { version = "1.41.0", optional = true }
 aws-config = { version = "1.5.5", optional = true }
-tokio-util = "0.7.12"
 
 [[example]]
 name = "github-tailer"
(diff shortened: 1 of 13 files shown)


runtime/plaid/Cargo.toml
```diff
@@ -56,6 +56,7 @@ tokio-tungstenite = { version = "0.23.1", features = [
 futures-util = "0.3.30"
 aws-sdk-kms = { version = "1.41.0", optional = true }
 aws-config = { version = "1.5.5", optional = true }
+tokio-util = "0.7.12"
```
obelisk, line 59: Put this with the other tokio imports.

runtime/plaid/src/benchmark/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// The full path to the output file where benchmarking metrics should be written
+    #[serde(default = "default_results_file_path")]
+    output_file_path: String,
+}
+
+/// Default file path for benchmarking results if none is provided in the config
+fn default_results_file_path() -> String {
+    format!(
+        "{}/../benchmark-results/metrics.txt",
+        env!("CARGO_MANIFEST_DIR")
```
obelisk, line 20: I think this is an absolute path which we don't want to pull into compilation as the binary could be moved anywhere later. I would just keep it relative in the same folder or in an adjacent subfolder.

runtime/plaid/src/executor/mod.rs
```diff
@@ -437,11 +451,29 @@ fn process_message_with_module(
                         format!("{}_computation_percentage_used", module.name),
                         computation_used as i64,
                     )?;
+
+                    // If benchmarking is enabled, log data to the benchmarking system
+                    if let Some(ref sender) = benchmark_mode {
+                        if let Err(e) = sender.blocking_send(ModulePerformanceMetadata {
```
obelisk, line 457: Why is this a blocking send when send everywhere else has been fine?

obelisk (changes requested):
Have a few comments. I would also say the way you've structured this is larger than benchmarks, you written a performance aggregator and perhaps should call it as such.

Then we can use that to write benchmarking harnesses.

runtime/plaid/src/executor/mod.rs
wbssbw: The channel we use for the benchmarking system is the `tokio::sync::mpsc` one instead of `crossbeam_channel`. Can switch to crossbeam and use `recv_timeout` in the benchmarking loop instead

runtime/plaid/src/benchmark/mod.rs
wbssbw: Yeah good call

obelisk (approved):

PR #61, opened 2024-11-23
Add Github API to check user org membership

timweri:
- Add Github API to check user org membership. Returns `true` if the user is part of the org and `false` if they are not part of the org.


runtime/plaid/src/apis/github/members.rs
```diff
@@ -0,0 +1,35 @@
...
+        let org = self.validate_org(request.get("org").ok_or(ApiError::BadRequest)?)?;
+        let user = self.validate_username(request.get("user").ok_or(ApiError::BadRequest)?)?;
+
+        info!("Checking if user [{user}] is part the [{org}] org on behalf of {module}");
+        let address = format!("/orgs/{org}/members/{user}");
+
+        match self.make_generic_get_request(address, module).await {
+            Ok((status, Ok(_))) => {
+                if status == 204 {
+                    Ok("true".to_string())
```
wbssbw, line 22: Any reason for having this return `String` instead of `bool`?

timweri: The macro cannot handle bool. I felt like it was easier to return String than making the macro more generic

timweri: Okay I was silly. Managed to use `bool` by using the right macro

obelisk (approved):

PR #63, opened 2024-11-27
Add github functionality to comment on a pull request

macalicious:
This Pull Request adds functionality to comment on a Pull Request based on the:

1. Organization
2. Repository Name
3. Pull Request Number

The comment is submitted as a string and the corresponding `{ "body": "<comment>" }` is constructed in the runtime.


runtime/plaid-stl/src/github/mod.rs
```diff
@@ -1545,12 +1548,48 @@ pub fn check_org_membership_of_user(
...
+    // There was an error with the Plaid system. Maybe the API is not
+    // configured.
+    Ok(res != 0)
+}
+
+pub fn comment_on_pull_request(
+    organization: &str,
+    repository_name: &str,
+    pull_request: &str,
+    comment: &str,
```
obelisk, line 1560: Make all of these `impl Display`

runtime/plaid/src/apis/github/repos.rs
```diff
@@ -316,4 +317,45 @@ impl Github {
...
+    /// See https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28#create-an-issue-comment for more detail
+    pub async fn comment_on_pull_request(
+        &self,
+        params: &str,
+        module: &str,
+    ) -> Result<u32, ApiError> {
+        let request: HashMap<&str, &str> =
+            serde_json::from_str(params).map_err(|_| ApiError::BadRequest)?;
+
+        let organization = self.validate_username(request.get("organization").ok_or(ApiError::BadRequest)?)?;
```
obelisk, line 331: Are you sure you want to call it organization instead of username?

runtime/plaid-stl/src/github/mod.rs
```diff
@@ -820,7 +823,7 @@ pub fn fetch_file(
     extern "C" {
         new_host_function_with_error_buffer!(github, fetch_file);
     }
-    const RETURN_BUFFER_SIZE: usize = 1024 * 1024; // 1 MiB
+    const RETURN_BUFFER_SIZE: usize = 10 * 1024 * 1024; // 10 MiB
```
obelisk, line 826: This will break a bunch of existing rules when they suddenly don't have enough memory to allocate this buffer

obelisk (changes requested):
Why do you want to 10x the buffer size. That is quite the dangerous proposition

obelisk (approved):

PR #65, opened 2024-12-05
Add tests

michelemin:
This PR adds the following tests:
* log back system (with delayed log back)
* reading from / writing to the DB
* opening a file on the host from a rule (impossible)
* get random bytes
* MNRs: not allowed, with URL templating, with dynamic headers

### DIFF
--- a/.github/workflows/Checks.yml
+++ b/.github/workflows/Checks.yml
@@ -14,14 +14,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - name: Get Rust Nightly
-      run: rustup toolchain install nightly
-
-    - name: Use Rust Nightly
-      run: rustup default nightly
-
-    - name: Get Wasm32
-      run: rustup target add wasm32-unknown-unknown
+    - name: Setup Rust
+      run: |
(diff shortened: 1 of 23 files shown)


testing/integration.sh
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
-set -e
```
michelemin, line 2: This is because, if we have a failing test, then the script would stop immediately at

```bash $module/harness/harness.sh```

Instead, we want the script to continue, since we have the mechanism to handle errors directly in the code.

File: .github/workflows/Checks.yml [hunk 2]
```diff
@@ -51,12 +53,13 @@ jobs:
           ~/.cargo/registry/index/
           ~/.cargo/registry/cache/
           ~/.cargo/git/db/
-          target/
-        key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          runtime/target/
+        key: ${{ runner.os }}-cargo-amd64-${{ hashFiles('**/Cargo.lock') }}
```
michelemin, line 57: These changes in the cache key are to avoid conflicts between caches for the different jobs.

obelisk (approved):

PR #71, opened 2025-01-07
Add size limit for storing in the DB

michelemin:
This PR adds an optional size limit that modules must respect when storing data in the permanent storage (e.g., DB).
If a module tries to store more than it is allowed, the insertion will fail.

Modules can also be given (by default or through overrides) `Unlimited` storage in the DB, in which case we will still count how many bytes they are storing, but we will not fail the insert.

When counting what is stored, we take into account the byte length of the key and the values.

### Breaking changes
The configuration needs a new section that looks like

```
[loading.storage_size]
default = "Unlimited"
[loading.storage_size.log_type]
[loading.storage_size.module_overrides]
"test_db.wasm" = { Limited = 50 }
```

### DIFF
--- a/runtime/plaid/src/functions/storage.rs
+++ b/runtime/plaid/src/functions/storage.rs
@@ -26,7 +26,10 @@ pub fn insert(
     let memory_view = match get_memory(&env, &store) {
         Ok(memory_view) => memory_view,
         Err(e) => {
-            error!("{}: Memory error in storage_insert: {:?}", env_data.name, e);
+            error!(
+                "{}: Memory error in storage_insert: {:?}",
+                env_data.module.name, e
+            );
             return FunctionErrors::CouldNotGetAdequateMemory as i32;
         }
     };
@@ -34,7 +37,10 @@ pub fn insert(
     let key = match safely_get_string(&memory_view, key_buf, key_buf_len) {
         Ok(s) => s,
         Err(e) => {
-            error!("{}: Key error in storage_insert: {:?}", env_data.name, e);
+            error!(
+                "{}: Key error in storage_insert: {:?}",
+                env_data.module.name, e
+            );
             return FunctionErrors::ParametersNotUtf8 as i32;
         }
     };
@@ -43,7 +49,10 @@ pub fn insert(
     let value = match safely_get_memory(&memory_view, value_buf, value_buf_len) {
         Ok(d) => d,
         Err(e) => {
-            error!("{}: Value error in storage_insert: {:?}", env_data.name, e);
+            error!(
+                "{}: Value error in storage_insert: {:?}",
+                env_data.module.name, e
+            );
             return FunctionErrors::CouldNotGetAdequateMemory as i32;
         }
     };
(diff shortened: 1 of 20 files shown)


runtime/plaid/src/functions/storage.rs
```diff
@@ -49,31 +49,88 @@ pub fn insert(
...
+        Err(_) => {
+            return FunctionErrors::InternalApiError as i32;
+        }
+    };
+
+    // Calculate the amount of storage that would be used after successfully inserting.
+    // Note: we _substract_ the size of existing data. If we were to insert the new data, the old data would be overwritten.
+    let used_storage = match env_data.storage_current.read() {
+        Ok(data) => *data,
+        Err(_) => panic!(),
```
obelisk, line 78: We should never panic. This needs to be surfaced and handled in some way.

obelisk (changes requested):

runtime/plaid/src/functions/storage.rs
```diff
@@ -43,43 +49,99 @@ pub fn insert(
...
+        },
+        Err(_) => {
+            return FunctionErrors::InternalApiError as i32;
+        }
+    };
+
+    // Calculate the amount of storage that would be used after successfully inserting.
+    // Note: we _substract_ the size of existing data. If we were to insert the new data, the old data would be overwritten.
+    let would_be_used_storage =
+        *storage_current + key_len as u64 + value.len() as u64 - existing_data_size; // no problem with underflowing because the result will never be negative (since used_storage >= existing_data_size)
```
obelisk, line 97: The comment should say `*storage_current >= existing_data_size`

runtime/plaid/src/functions/storage.rs
```diff
@@ -43,43 +49,99 @@ pub fn insert(
...
+
+    // If we have limited storage, this insert might fail
+    if let LimitValue::Limited(storage_limit) = env_data.module.storage_limit {
```
obelisk, line 100: Will this still work if a module has its storage size reduced to below what is currently used?

runtime/plaid/src/functions/storage.rs
```diff
@@ -43,43 +49,99 @@ pub fn insert(
...
+            error!("Critical error getting a lock on used storage: {:?}", e);
+            return FunctionErrors::InternalApiError as i32;
+        }
+    };
+
+    // We check if this insert would overwrite some existing data. If so, we need to take that into account when
+    // computing the storage that would be occupied at the end of the insert operation.
+    // Note: if we have existing data, then we need to count the key's length as well. This is because at the end
+    // of a possible insertion, we would have only one key.
+    let existing_data_size = match env_data
```
obelisk, line 79: This doesn't need the lock so do it before acquiring it. That will increase parallelism

obelisk (changes requested):

runtime/plaid/src/functions/storage.rs
timweri: Are you talking about fetching `existing_data_size`? Wouldn't there be a race condition if there's another instance of the same module updating the same key?

runtime/plaid/src/functions/storage.rs
michelemin: After discussing this, the conclusion is that yes, it would still work: all inserts would fail until a sufficient number of bytes is deleted from the DB (thus bringing the counter down).

runtime/plaid/src/functions/storage.rs
michelemin: Yes, we discussed this and several other alternatives because of concerns on performance, but there were always caveats around concurrency. For now, the decision was to
* keep thinking about the optimal solution
* restructure the code so that we pay no performance penalties (no locks, etc.) in case of `Unlimited` storage

obelisk (approved):

runtime/plaid/src/functions/storage.rs
obelisk: Correct, I was wrong.

PR #73, opened 2025-01-14
Secrets manager

michelemin:
This PR adds a Rust tool to easily manage Plaid secrets. The tool allows one to
* Read a file with Plaid secrets and upload them to AWS Secrets Manager, with appropriate names.
* Fetch secrets from AWS Secrets Manager and assemble them in a file that Plaid can consume.

We also add a CI that releases the tool's binary as a GH release whenever a tag is pushed.

Moreover, this PR redesign how Plaid is released: a main workflow (`Release.yaml`) calls two sub-workflows that upload artifacts; the main workflow then collects these artifacts and makes the release.
Without these changes, the two workflows would have tried to create a release with the same name. And we don't even want to manage multiple releases: we want one release with all the artifacts.


runtime/plaid/src/bin/secrets_manager.rs
```diff
@@ -0,0 +1,240 @@
...
+fn secret_name_to_json_name(instance: impl Display, s: impl Display) -> String {
+    let input = s.to_string();
+    let stripped = input.strip_prefix(&format!("plaid-{instance}-")).unwrap();
+    format!("{{plaid-secret{{{stripped}}}}}")
+}
+
+/// Parse the CLI arguments
+fn parse_args() -> Options {
+    let matches = Command::new("Plaid Secrets Manager")
+        .version("0.1.0")
```
obelisk, line 52: Use the package version. That means it'll be much easier to know what we're working with when the binary inevitably gets separated from the rest of the repository.

runtime/plaid/src/bin/secrets_manager.rs
```diff
@@ -0,0 +1,240 @@
...
+        .about("A simple tool that helps with managing Plaid secrets")
+        .arg(Arg::new("instance")
+            .long("instance")
+            .help("Specifies the type of instance (plaid or ingress)")
+            .required(true)
+            .value_parser(["plaid", "ingress"])
```
obelisk, line 58: You sure you want to be this prescriptive? In the future there might be different ones and should the secrets parser care?

runtime/plaid/src/bin/secrets_manager.rs
```diff
@@ -0,0 +1,240 @@
...
+                .args(["json_to_aws", "aws_to_json"])
+                .multiple(false)
+                .required(true)
+        )
+        .get_matches();
+
+    let instance = matches.get_one::<String>("instance").unwrap().to_string(); // unwrap OK because the param is required
+    let region = matches
+        .get_one::<String>("region")
+        .unwrap_or(&"us-east-1".to_string())
```
obelisk, line 88: Take a reference to a static value you just turned into a string?

runtime/plaid/src/bin/secrets_manager.rs
```diff
@@ -0,0 +1,240 @@
...
+    // Read and parse the file's content
+    let contents = std::fs::read_to_string(filename.to_string()).unwrap();
+    let value = serde_json::from_str::<Value>(&contents).unwrap();
+    let value = value.as_object().unwrap();
+
+    // Fill a vector with all the secrets, ready to be uploaded
+    let mut secrets = vec![];
+    for (key, value) in value {
+        secrets.push(PlaidSecret {
+            name: json_name_to_secret_name(&secret_name_regex, instance.to_string().clone(), key),
```
obelisk, line 121: Unneeded clone?

runtime/plaid/src/bin/secrets_manager.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    Options {
+        instance,
+        region,
+        operation,
+    }
+}
+
+/// Read a file with Plaid secrets and upload them to AWS Secrets Manager, with appropriate names.
+async fn json_to_aws(filename: impl Display, instance: impl Display, sm_client: &Client) {
```
obelisk, line 109: `AsRef<str>` might be what you want

runtime/plaid/src/bin/secrets_manager.rs
```diff
@@ -0,0 +1,240 @@
...
+    let contents = std::fs::read_to_string(filename.to_string()).unwrap();
+    let value = serde_json::from_str::<Value>(&contents).unwrap();
+    let value = value.as_object().unwrap();
+
+    // Fill a vector with all the secrets, ready to be uploaded
+    let mut secrets = vec![];
+    for (key, value) in value {
+        secrets.push(PlaidSecret {
+            name: json_name_to_secret_name(&secret_name_regex, instance.to_string().clone(), key),
+            value: value.as_str().unwrap().to_string(),
```
obelisk, line 122: Seems dangerous. Should there be some pre-processing?

runtime/plaid/src/bin/secrets_manager.rs
```diff
@@ -0,0 +1,240 @@
...
+        });
+    }
+
+    // Upload secrets to SM
+    for secret in secrets {
+        println!("Uploading {}...", secret.name);
+        match sm_client
+            .create_secret()
+            .name(secret.name.clone())
+            .kms_key_id("alias/plaid-dev-encrypt-decrypt")
```
obelisk, line 132: Hard coding dev?

runtime/plaid/src/bin/secrets_manager.rs
```diff
@@ -0,0 +1,240 @@
...
+                }
+            }
+        }
+    }
+}
+
+/// Fetch secrets from AWS Secrets Manager and assemble them in a file that Plaid can consume
+async fn aws_to_json(filename: impl Display, instance: impl Display, sm_client: &Client) {
+    println!("Fetching all secrets whose name starts with plaid-{instance}");
+    let mut next_token: Option<String> = Some("".to_string());
```
obelisk, line 158: String::new

obelisk: Also I don't think you need to specify the type explicitly

runtime/plaid/src/bin/secrets_manager.rs
```diff
@@ -0,0 +1,240 @@
...
+    let mut first_request = true;
```
obelisk, line 159: This is a code smell. Feels like tracking state this way is going to be difficult to maintain

runtime/plaid/src/bin/secrets_manager.rs
```diff
@@ -0,0 +1,240 @@
...
+    let mut retrieved_secrets = vec![];
+
+    while next_token.is_some() {
+        // The first request is special: we want next_token to be None
+        if first_request {
+            first_request = false;
+            next_token = None;
+        }
```
obelisk, line 167: This seems like something to do outside the loop

obelisk (changes requested):
There are a lot of unwraps here which concern me.

I also think the state machine for pulling secrets should be revised.

runtime/plaid/Cargo.toml
```diff
@@ -12,8 +12,9 @@ aws = ["aws-sdk-kms", "aws-config"]
 [dependencies]
 alkali = "0.3.0"
 async-trait = "0.1.56"
+aws-sdk-secretsmanager = "1.57.0"
 base64 = "0.13"
-clap = { version = "4", default-features = false, features = ["std"] }
+clap = { version = "4", features = ["std"] }
```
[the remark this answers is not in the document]
michelemin: This is to have things like help and usage. If you prefer, we can try selecting these features one by one, or leave them out altogether and have no help messages.

runtime/plaid/src/bin/secrets_manager.rs
```diff
@@ -0,0 +1,240 @@
...
+    AwsToJson(String),
+}
+
+struct Options {
+    instance: String,
+    region: String,
+    operation: Operation,
+}
+
+struct PlaidSecret {
```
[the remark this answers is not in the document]
michelemin: I put it here because my aim, at least for now is not to change anything in the runtime lib. Meaning that Plaid will continue seeing a secrets.json file, which just happens to be constructed in a different way.
If we go for a deeper rewrite of the secrets system, we can consider sharing this struct and then doing something in the runtime, but I think that will take longer.

runtime/plaid/src/bin/secrets_manager.rs
```diff
@@ -0,0 +1,240 @@
...
+    value: String,
+}
+
+/// Take something that looks like `{plaid-secret{secret-name}}`
+/// and extract `secret-name`. Then prepend it with `plaid-plaid-` or `plaid-ingress-`
+/// depending on which instance we are processing for.
+fn json_name_to_secret_name(r: &regex::Regex, instance: impl Display, s: impl Display) -> String {
+    let inner_name = r
+        .captures(&s.to_string())
+        .unwrap()
```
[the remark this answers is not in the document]
michelemin: No, I think most of these unwraps are not safe. I was going for the "if it does not work, it's unrecoverable anyway, so panic" approach, assuming everything needs to be fixed by hand, before rerunning the tool.

runtime/plaid/src/bin/secrets_manager.rs
michelemin: Good idea!

runtime/plaid/src/bin/secrets_manager.rs
michelemin: Fair point. This is just to validate that we don't have things like `ingres`, `plald`, etc. because that would mess up the secrets' names in Secrets Manager. So it's a trade-off, but we can leave out this validation if you prefer.

runtime/plaid/src/bin/secrets_manager.rs
michelemin: Haha now that you point it out...... I think I was in "please just compile" mode

runtime/plaid/src/bin/secrets_manager.rs
michelemin: Well spotted! I will make it configurable in the CLI params

runtime/plaid/src/bin/secrets_manager.rs
michelemin: Yeah, it does not look good. I basically need a do-while, so I can try with this: https://users.rust-lang.org/t/do-while-in-rust-language/39361

runtime/plaid/src/bin/secrets_manager.rs
michelemin: Ended up using a default value in the CLI arg, which is cleaner.

runtime/plaid/src/bin/secrets_manager.rs
michelemin: I added a note to say the method will panic if the file contains unexpected data. Let's discuss whether this is sufficient for what we need right now from this tool.

obelisk (approved):

PR #74, opened 2025-01-21
Split get_accessory_data to avoid mixing secrets, headers, etc.

michelemin:
The current situation is that secrets, headers (for POST requests) and query params (for GET requests) end up in "accessory data".
The code avoids security issues by making sure that secrets are added last, thus overwriting headers/params with the same name. However, this remains error prone and prevents us from having things with the same name.

This PR splits this additional data into
* generic "accessory data" (with "universal" accessory data available to all rules, plus overrides based on log type and file name)
* secrets
* headers
* query parameters

Each has their getter method, generated through macros (since it's the same exact thing for all of them).

Also adding tests that show that
* the basic mechanism works
* having secrets with the same name as headers / query params works
* universal accessory data and its override mechanisms work as expected, with `filename overrides > logtype overrides > universal accessory data`, where `>` means "takes precedence over".

Existing tests have been updated.

# Breaking changes
! Existing rules need to be changed !
Usually rules are getting either secrets or headers from accessory data. One has to go through all the rules and replace that call with `get_headers` or `get_secrets`, as appropriate.

Also, the configuration will need a new section like

```
[loading.universal_accessory_data]
"key_1" = "value_1"
"key_2" = "value_2"
"key_4" = "value_4_ignored"

[loading.accessory_data_log_type_overrides.test_geteverything]
"key_1" = "value_1_intermediate"
"key_4" = "value_4"

[loading.accessory_data_file_overrides."test_get_everything.wasm"]
"key_1" = "value_1_new"
"key_3" = "value_3"
```


runtime/plaid-stl/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "plaid_stl"
-version = "0.14.2"
+version = "0.16.0"
```
obelisk, line 3: Currently would be 15

runtime/plaid-stl/src/plaid/mod.rs
```diff
@@ -100,59 +100,70 @@ pub fn get_time() -> u32 {
...
-    let name_bytes = name.as_bytes().to_vec();
-
-    let buffer_size = unsafe {
-        fetch_accessory_data_by_name(
-            name_bytes.as_ptr(),
-            name_bytes.len(),
-            vec![].as_mut_ptr(),
-            0,
-        )
+macro_rules! generate_getter {
```
obelisk, line 103: Perhaps a more descriptive name? `generate_string_getter`?

runtime/plaid-stl/src/plaid/mod.rs
```diff
@@ -100,59 +100,70 @@ pub fn get_time() -> u32 {
...
-        return Err(PlaidFunctionError::InternalApiError);
-    }
-
-    match String::from_utf8(data_buffer) {
-        Ok(s) => Ok(s),
-        Err(_) => Err(PlaidFunctionError::ParametersNotUtf8),
-    }
 }
 
+generate_getter!(accessory_data);
```
obelisk, line 162: Documentation

runtime/plaid/src/bin/plaid.rs
```diff
@@ -260,7 +260,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                                 let message = Message {
                                     type_: name.to_string(),
                                     data: String::new().into_bytes(),
-                                    accessory_data: query.into_iter().map(|(k, v)| (k, v.into_bytes())).collect(),
+                                    accessory_data: HashMap::new(),
```
obelisk, line 263: Does this break deserialization of previous messages?

runtime/plaid/src/data/interval/mod.rs
```diff
@@ -14,6 +14,7 @@ use crate::executor::Message;
 /// Defines the list of interval jobs to be processed
 pub struct IntervalConfig {
     /// A HashMap of job name to job config. The job's name will be included in accessory data as "job_name"
+    /// TODO check if the job name is really included in accessory data. Could not find where.
```
obelisk, line 17: Resolve before merging. @wbssbw 

I don't think it is. I think the job name is included in log_source metadata

runtime/plaid/src/functions/message.rs
```diff
@@ -133,57 +133,68 @@ pub fn fetch_source(env: FunctionEnvMut<Env>, data_buffer: WasmPtr<u8>, buffer_s
...
         }
-    } else {
-        // If there is no field with that name, we return 0 similar to
-        // fetching the from_module
-        0
-    }
+    };
 }
+
+generate_getter!(accessory_data);
```
obelisk, line 197: Documentation

obelisk: Or a description of where documentation can be found.

obelisk (changes requested):

runtime/plaid-stl/Cargo.toml
michelemin: Yes, let's see if we merge this one of the DB size limit first.

runtime/plaid/src/bin/plaid.rs
michelemin: I don't think so. In the sense that whatever is given as query params ends up in the `query_params` field of the `Message`, while the `accessory_data` remains empty (for now). Then it's up to the rule to know that data has to be read differently. So yes, in that sense this is a breaking change. And same for `secrets` and `headers`.

runtime/plaid/src/data/interval/mod.rs
wbssbw: @obelisk is correct. At one point you could fetch the interval job's name from accessory data using "job_name" as the key, but that has since changed. The job name defined here is used in `LogSource::Generator(Generator::Interval(name))`

michelemin: OK thanks for checking! I will remove those two lines.

runtime/plaid/src/loader/mod.rs
```diff
@@ -365,6 +375,28 @@ pub async fn load(
...
+            // If we already had accessory data, start from there. Otherwise, start from an empty map
+            let mut accessory_data = match plaid_module.accessory_data {
+                Some(ref v) => v.clone(),
+                None => HashMap::new(),
+            };
+            for (key, value) in prad {
+                accessory_data.insert(key.to_string(), value.as_bytes().to_vec());
+            }
+            plaid_module.accessory_data = Some(accessory_data);
+        }
```
michelemin, line 399: @obelisk I have a feeling this part can be improved

runtime/plaid/src/loader/mod.rs
```diff
@@ -365,6 +375,28 @@ pub async fn load(
...
```
michelemin, line 399: @obelisk I have a feeling this part can be improved. Do you have any suggestions?

obelisk:
@michelemin Can you resolve conflicts so I can merge this?

michelemin:
> @michelemin Can you resolve conflicts so I can merge this?

Done!

obelisk (approved):

PR #76, opened 2025-02-06
Okta logs: ensure we do not miss any logs

michelemin:
This PR does the following:
* Make the sorting of Okta logs configurable. Changed the default to `ASCENDING`, but let's discuss that.
* Introduce a data structure (LRU cache) where we store the UUIDs of the logs that we have already seen and processed.
* Query for logs since `last_seen - 1_minute`, to have overlapping queries and ensure (with higher probability) we do not miss logs.

The whole problem comes from the fact that Okta does not guarantee logs to be ordered (see [here](https://support.okta.com/help/s/question/0D54z000070QfVwCAK/api-getting-okta-system-logs-sometimes-updates-are-missing?language=en_US) ), so if we query for logs `since` the last seen event, we might miss events that happened before but are not yet available through the API. By subtracting 1 minute, we try to be more resilient and capture all events that might have appeared in the meantime.

With this updated mechanism, we will make overlapping queries and potentially see the same logs more than once. When we process a log, we add its UUID to an LRU cache with capacity N, meaning it will store the identifiers for the last N logs we have sent on to Plaid's rules. When the cache is full and a new item is added, the least-recently-used is automatically removed.

Note - We do not want to process the same log multiple times. The only case where this can happen is across Plaid restarts, where we would go 1 minute back in time and risk processing logs that had already been processed before the reboot. Persisting the cache across reboots would fix this problem, but seems an overkill.

Note2 - If an event appears more than 1 minute after happening, we will still miss it because we won't go back in time enough. We can tweak the "1 minute" parameter to achieve the desired trade-off.


obelisk (changes requested):
While this works, it doesn't guarantee that logs are delivered in time order (as opposed to the GitHub Audit logger which uses a heap).

Are we sure this is ok?

obelisk (approved):
We chatted about this and since I was originally going through in descending order anyway logs were already out of order.

Thus we're going to merge this to fix the DG, then we will merge the GHAL and this to share the same codebase.

PR #77, opened 2025-02-07
Fix bug in Okta DG

michelemin:
Okta DG could enter a deadlock if we get more than 100 logs in one minute. More detailed explanation in the comment to the fix.


obelisk (approved):
I don't like it but all previous ones are broken and this might not be

PR #79, opened 2025-02-24
Unify data generators for GH and Okta

michelemin:
This PR unifies the logic for data generators (Okta and GitHub).

The basic idea is that each of Okta and GitHub DG implement a specific logic to fetch some logs from the respective API that were created in a given time window (since...until).

Then there is higher-level logic to take these logs and process them, discarding duplicates, etc. This logic is common between Okta and GH.

Commonalities are expressed through a `DataGenerator` trait which both sources implement. Methods from this trait are then called from the high-level logic to perform all necessary operations.

### DIFF
--- a/runtime/plaid/src/data/mod.rs
+++ b/runtime/plaid/src/data/mod.rs
@@ -192,7 +192,7 @@ struct DataGeneratorLog {
 
 /// This is what data generators have in common
 #[allow(async_fn_in_trait)]
-trait DataGenerator {
+pub trait DataGenerator {
     /// Fetch from the source some logs that were produced between `since` and `until`.
     ///
     /// Note: this function does not necessarily return _all_ logs that were produced in that time frame,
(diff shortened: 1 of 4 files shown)


runtime/plaid/src/data/mod.rs
```diff
@@ -110,8 +115,8 @@ impl Data {
         if let Some(mut gh) = di.github {
             handle.spawn(async move {
                 loop {
-                    if let Err(e) = gh.fetch_audit_logs().await {
-                        error!("GitHub Data Fetch Error: {}", e)
+                    if let Err(_) = big_logic(&mut gh).await {
```
michelemin, line 118: Note to self: rename this 😄

runtime/plaid/src/data/mod.rs
```diff
@@ -174,3 +179,161 @@ impl Data {
...
+    ) -> Result<Vec<DataGeneratorLog>, ()>;
+
+    fn get_name(&self) -> String;
+    fn get_sleep_duration(&self) -> u64;
+    fn get_canon_time(&self) -> u64;
+    fn get_last_seen(&self) -> OffsetDateTime;
+    fn set_last_seen(&mut self, v: OffsetDateTime);
+    fn was_already_seen(&self, id: impl Display) -> bool;
+    fn mark_already_seen(&mut self, id: impl Display);
+    fn send_for_processing(&self, payload: Vec<u8>);
```
michelemin, line 212: Note: add doc comments

obelisk (commented):
Run tests?

michelemin:
@obelisk They ran but they don't appear in the tab at the top of this PR: https://github.com/obelisk/plaid/actions/runs/13525709200
That's strange 🤔

obelisk (approved):

PR #80, opened 2025-02-25
Testing mode

obelisk:
Add a way to prevent modules from having side effects at the runtime level. This allows you to run modules where all calls that could have side effects will be disabled by the runtime; allowing more safe testing of rules before production deployments.

Log backs, storage, cache, and Slack are exempt from these rules even though they have side effects.

MNRs are configurable on a case by case basis defaulting to not being allowed.

### DIFF
--- a/runtime/plaid/src/apis/general/network.rs
+++ b/runtime/plaid/src/apis/general/network.rs
@@ -1,5 +1,6 @@
-use std::collections::HashMap;
+use std::{collections::HashMap, sync::Arc};
 
+use crate::loader::PlaidModule;
 use reqwest::header::HeaderMap;
 use serde::{Deserialize, Serialize};
 
@@ -22,8 +23,8 @@ struct MakeRequestRequest {
     /// Variables to include in the request. Variables take the place of an idenfitifer in the request URI
     variables: HashMap<String, String>,
     /// Dynamic headers to include in the request. These are headers that cannot be statically
-    /// defined in the request configuration. They cannot override a request's statically defined headers 
-    headers: Option<HashMap<String, String>>
+    /// defined in the request configuration. They cannot override a request's statically defined headers
+    headers: Option<HashMap<String, String>>,
 }
 
 /// This struct represents a web request and contains information about what the request is about (e.g., verb and URI),
@@ -44,6 +45,10 @@ pub struct Request {
     allowed_rules: Vec<String>,
     /// Headers to include in the request
     headers: HashMap<String, String>,
+    /// Whether the request is available to modules in test mode. Generally this should be set to false
+    /// if the call has side effects. If it is not set, this will default to false.
+    #[serde(default)]
+    available_in_test_mode: bool,
 }
 
 /// Data returned by a request.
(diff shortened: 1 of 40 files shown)


runtime/plaid/src/apis/general/network.rs
```diff
@@ -100,24 +120,37 @@ impl General {
...
             error!("{module} tried to use web-request which it's not allowed to: {request_name}");
             return Err(ApiError::BadRequest);
         }
 
+        // If the call is coming from a module in test mode, and the request is not allowed to be
+        // called in test mode, return TestMode error
+        if module.test_mode && !request_specification.available_in_test_mode {
+            error!("{module} tried to use web-request which is not available in test mode: {request_name}");
+            return Err(ApiError::TestMode);
+        }
```
michelemin, line 136: Just noticing that for MNRs we do the check here, while for `simple_json_post_request` we don't. Maybe we can follow the same approach?

runtime/plaid/src/functions/api.rs
```diff
@@ -257,86 +287,99 @@ macro_rules! impl_new_sub_module_function_with_error_buffer {
...
 
 // AWS functions
 #[cfg(feature = "aws")]
-impl_new_sub_module_function_with_error_buffer!(aws, kms, sign_arbitrary_message);
+impl_new_sub_module_function_with_error_buffer!(
+    aws,
+    kms,
+    sign_arbitrary_message,
+    DISALLOW_IN_TEST_MODE
+);
```
michelemin, line 334: Super-nit: I guess this could be allowed because it doesn't really cause any changes to external systems. I can see a case where a rule wants to get a message signed and check if the signature would be correct or something.

runtime/plaid/src/apis/general/network.rs
michelemin: OK I think I get it: it's because `simple_json_post_request` is always denied, while MNRs can have exemptions, right?

runtime/plaid/src/functions/api.rs
```diff
@@ -257,86 +287,99 @@ macro_rules! impl_new_sub_module_function_with_error_buffer {
 }
 
 // General Functions
-impl_new_function!(general, simple_json_post_request);
-impl_new_function_with_error_buffer!(general, make_named_request);
+impl_new_function!(general, simple_json_post_request, DISALLOW_IN_TEST_MODE);
+impl_new_function_with_error_buffer!(general, make_named_request, ALLOW_IN_TEST_MODE);
```
michelemin, line 291: Just to check I get everything: here we allow all MNRs and then when it's time to make the call we check, so that the MNR goes through only if that specific request is added to the exemption list.

michelemin (approved):

runtime/plaid/src/apis/general/network.rs
obelisk: That's correct. MNRs are denied by default but you can add a flag to say its allowed. Where as simple_json_post is too expressive and should be denied

runtime/plaid/src/functions/api.rs
obelisk: I struggled with this as well as the rustica one but eventually decided against it because those signatures could allow access outside the Plaid system (Rustica obviously does). So while this doesn't mutate anything by letter of the law, it does spiritually.

runtime/plaid/src/functions/api.rs
obelisk: This implementation allows multiple levels of checks. If you're writing a new API that always has side effects, you can just specify here `DISALLOW_IN_TEST_MODE` and you'll never have to think about it in the implementation.

But you might have an API that only **sometimes** has side effects, like MNRs. This allows you to say the API is available in test mode, but then your implementation can do extra checks. If you then decide not to allow the call, you can return `ApiError::TestMode` which will get converted to `FunctionError::TestMode` by the macro. Eventually this becomes passed to the rule as error `-11` so it knows that what it called was not allowed because its running in test mode.

runtime/plaid/src/functions/api.rs
michelemin: Makes sense

PR #81, opened 2025-02-28
Support DBs shared between rules

michelemin:
Sometimes it makes sense for rules to share a common database, so that data written by a rule can be read by another one. For example, this can remove the need for log-backs in case we need to pass information from rule A to rule B and the synchronization is already handled somewhere else.

This PR introduces support for shared DBs in a fully backward-compatible way. We add new APIs like `insert_shared`, `get_shared`, `delete_shared` that also take a namespace, which represents the shared DB.
Moreover, at the config level, we can specify which rules have read-only access and which rules can read/write.
Note: write-only is not currently supported because a write op inherently requires a read op to fetch existing data under the same key. This could change in the future.

### DIFF
--- a/runtime/plaid/src/functions/storage.rs
+++ b/runtime/plaid/src/functions/storage.rs
@@ -6,6 +6,36 @@ use crate::{executor::Env, functions::FunctionErrors, loader::LimitValue, storag
 
 use super::{get_memory, safely_get_memory, safely_get_string, safely_write_data_back};
 
+macro_rules! safely_get_guest_string {
+    ($variable:ident, $memory_view:expr, $buf:expr, $buf_len:expr, $env_data:expr) => {
+        let $variable = match safely_get_string(&$memory_view, $buf, $buf_len) {
+            Ok(s) => s,
+            Err(e) => {
+                error!(
+                    "{}: error while getting a string from guest memory: {:?}",
+                    $env_data.module.name, e
+                );
+                return FunctionErrors::ParametersNotUtf8 as i32;
+            }
+        };
+    };
+}
+
+macro_rules! safely_get_guest_memory {
+    ($variable:ident, $memory_view:expr, $buf:expr, $buf_len:expr, $env_data:expr) => {
+        let $variable = match safely_get_memory(&$memory_view, $buf, $buf_len) {
+            Ok(d) => d,
+            Err(e) => {
+                error!(
+                    "{}: error while getting bytes from guest memory: {:?}",
+                    $env_data.module.name, e
+                );
+                return FunctionErrors::ParametersNotUtf8 as i32;
+            }
+        };
+    };
+}
+
 /// Code which is common to `insert` and `insert_shared`
 fn insert_common(
     env_data: &Env,
@@ -77,6 +107,11 @@ fn insert_common(
             // If we would go above the limited storage, reject the insert
             if would_be_used_storage > storage_limit {
                 error!("{}: Could not insert key/value with key {key} as that would bring us above the configured storage limit.", env_data.module.name);
+                let _ = env_data.external_logging_system.log_module_error(
+                    env_data.module.name.clone(),
+                    "Could not insert key/value as that would bring us above the configured storage limit.".to_string(),
+                    vec![]
+                );
                 return FunctionErrors::StorageLimitReached as i32;
             }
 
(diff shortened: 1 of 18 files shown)


runtime/plaid/src/functions/storage.rs
```diff
@@ -229,6 +343,145 @@ pub fn get(
...
+                "{}: Key error in storage_get_shared: {:?}",
+                env_data.module.name, e
+            );
+            return FunctionErrors::ParametersNotUtf8 as i32;
+        }
+    };
+
+    // Check if the rule can read from the shared DB.
+    // This is the case if the rule is in either the "r" or "rw" section of the config.
+    let allowed = match &storage.shared_dbs {
```
obelisk, line 450: I would split this check up because we can move if this is `None` to the top of function, before any memory operations.

runtime/plaid/src/functions/storage.rs
```diff
@@ -229,6 +343,145 @@ pub fn get(
...
+        None => false,
+        Some(dbs) => match dbs.get(&namespace) {
+            None => false,
+            Some(db) => {
+                db.config
+                    .r
+                    .as_ref()
+                    .unwrap_or(&Vec::new())
```
obelisk, line 458: Don't have these fields be optional in this way. Have them define `default` so if they're not provided they're already empty. Then you don't need to `unwrap_or_default` every time

runtime/plaid/src/functions/storage.rs
```diff
@@ -112,15 +76,14 @@ pub fn insert(
 
             // If we would go above the limited storage, reject the insert
             if would_be_used_storage > storage_limit {
-                error!("{}: Could not insert key/value with key {storage_key} as that would bring us above the configured storage limit.", env_data.module.name);
+                error!("{}: Could not insert key/value with key {key} as that would bring us above the configured storage limit.", env_data.module.name);
```
obelisk, line 79: Is it possible to also send this to els?

obelisk: Without the value

obelisk (changes requested):

runtime/plaid/src/functions/storage.rs
michelemin: Good point!

runtime/plaid/src/functions/storage.rs
```diff
@@ -419,3 +603,132 @@ pub fn delete(
...
+        None => false,
+        Some(db) => db.config.rw.contains(&env_data.module.name),
+    };
+    if !allowed {
+        return FunctionErrors::OperationNotAllowed as i32;
+    }
+
+    if !allowed {
+        return FunctionErrors::ApiNotConfigured as i32; // TODO better error?
+    }
```
obelisk, line 708: wat

obelisk (changes requested):

runtime/plaid/src/functions/storage.rs
michelemin: goto fail

runtime/plaid/src/functions/storage.rs
```diff
@@ -251,23 +354,128 @@ pub fn list_keys(
...
+    env: FunctionEnvMut<Env>,
+    prefix_buf: WasmPtr<u8>,
+    prefix_buf_len: u32,
+    data_buffer: WasmPtr<u8>,
+    data_buffer_len: u32,
+) -> i32 {
+    let store = env.as_store_ref();
+    let env_data = env.data();
+
+    let storage = if let Some(storage) = &env_data.storage {
```
obelisk, line 460: `ok_or`

runtime/plaid/src/functions/storage.rs
```diff
@@ -419,3 +603,128 @@ pub fn delete(
...
+    env: FunctionEnvMut<Env>,
+    key_buf: WasmPtr<u8>,
+    key_buf_len: u32,
+    data_buffer: WasmPtr<u8>,
+    data_buffer_len: u32,
+) -> i32 {
+    let store = env.as_store_ref();
+    let env_data = env.data();
+
+    let storage = if let Some(storage) = &env_data.storage {
```
obelisk, line 618: `ok_or`

runtime/plaid/src/storage/mod.rs
```diff
@@ -77,13 +114,45 @@ pub trait StorageProvider {
...
 impl Storage {
-    pub fn new(config: Config) -> Result<Self, StorageError> {
+    pub async fn new(config: Config) -> Result<Self, StorageError> {
         let database = match config.sled {
             Some(sled) => Box::new(sled::Sled::new(sled)?),
             _ => return Err(StorageError::NoStorageConfigured),
         };
 
-        Ok(Storage { database })
+        let shared_dbs = match config.shared_dbs {
```
obelisk, line 123: Could do this in a `map`

obelisk (dismissed):
Just a few nits

obelisk (approved):

PR #83, opened 2025-03-11
Add API to list keys from shared DB

michelemin:
Add API to list keys from shared DB


modules/tests/test_shared_db_rule_2/src/lib.rs
```diff
@@ -20,9 +20,16 @@ fn main(log: String, _: LogSource) -> Result<(), i32> {
             if r.len() != 2 {
                 panic!()
             }
+
+            plaid::print_debug_string(&format!("[{RULE_NAME}] Listing keys from DB..."));
+            let x = plaid::storage::list_keys_shared(SHARED_DB, None::<String>).unwrap();
```
obelisk, line 25: This is an odd `None`. Do you really need this strong typing?

obelisk (approved):

PR #85, opened 2025-03-18
Support AWS DynamoDB for Plaid DB

michelemin:
This PR adds support for AWS DynamoDB as Plaid's DB layer.

Some points to note:
* DynamoDB is very fast but still slower than a local DB on file. This is why we had to add `sleep ...` in the DB test, which would otherwise fail for ~unclear reasons. To be investigated more thoroughly, but this is likely due to network latency or consistency.~ reasons explained in the comment below.
* The mechanism to authenticate to AWS (either via API key or IAM role) has been refactored out of the KMS submodule, since it's generic enough to be reused. This does not prevent KMS and DynamoDB from using different authentications.
* We could consider gating DynamoDB support behind a feature with optional dependencies.
* ~There is a TODO in the code for further consideration. When fetching all the keys, we currently fetch the entire namespace (keys and values) and then discard the values. This works but is suboptimal.~ Fixed with some DynamoDB magic ✅ 

Open tasks:
- [x] Investigate performance (see point above about `sleep`)
- [x] Use constants instead of hardcoded strings for "namespace", "key", "value"
- [x] Write more tests
  - [x] Listing keys or key/value pairs _with prefix_
  - [x] Inserting on top of an existing value returns the old value
  - [x] Deleting an existing value returns the old value
- [x] Put DynamoDB behind an `aws` feature together with KMS
- [x] Add tests to verify everything compiles with and without `aws` feature

### Config

The format for the DB config becomes as follows:
```
[storage.db]
sled_path = "/tmp/sled"

# [storage.db]
# table_name = "test-plaid"

# [storage.db.authentication]
# access_key_id = "value here"
# secret_access_key = "value here"
# session_token = "value here"
# region = "value here"
```
where one sets either Sled (with the path) or DynamoDB (with table name and auth) as the DB layer.

### DIFF
--- a/runtime/plaid/Cargo.toml
+++ b/runtime/plaid/Cargo.toml
@@ -12,6 +12,9 @@ aws = ["aws-sdk-kms", "aws-config"]
 [dependencies]
 alkali = "0.3.0"
 async-trait = "0.1.56"
+aws-config = { version = "1.5.5", optional = true }
+aws-sdk-dynamodb = "1.69.0"
+aws-sdk-kms = { version = "1.41.0", optional = true }
 aws-sdk-secretsmanager = "1.57.0"
 base64 = "0.13"
 clap = { version = "4", default-features = false, features = [
@@ -22,8 +25,10 @@ clap = { version = "4", default-features = false, features = [
 crossbeam-channel = "0.5"
 env_logger = "0.8"
 flate2 = "1.0"
+futures-util = "0.3.30"
 hex = "0.4.3"
 http = "1"
+jsonwebtoken = { version = "9.2" }
 jwt-simple = { version = "0.12.10", default-features = false, features = [
     "pure-rust",
 ] }
@@ -58,10 +63,6 @@ urlencoding = "2.1.3"
 warp = { version = "0.3", features = ["tls"] }
 wasmer = { version = "4", default-features = false, features = ["cranelift"] }
 wasmer-middlewares = "4"
-jsonwebtoken = { version = "9.2" }
-futures-util = "0.3.30"
-aws-sdk-kms = { version = "1.41.0", optional = true }
-aws-config = { version = "1.5.5", optional = true }
 
 [[example]]
 name = "github-tailer"
--- a/runtime/plaid/Cargo.toml
+++ b/runtime/plaid/Cargo.toml
@@ -1,19 +1,19 @@
 [package]
 name = "plaid"
-version = "0.19.0"
+version = "0.20.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
 default = ["aws"]
-aws = ["aws-sdk-kms", "aws-config"]
+aws = ["aws-sdk-kms", "aws-sdk-dynamodb"]
 
 [dependencies]
 alkali = "0.3.0"
 async-trait = "0.1.56"
-aws-config = { version = "1.5.5", optional = true }
-aws-sdk-dynamodb = "1.69.0"
+aws-config = "1.5.5"
+aws-sdk-dynamodb = { version = "1.69.0", optional = true }
 aws-sdk-kms = { version = "1.41.0", optional = true }
 aws-sdk-secretsmanager = "1.57.0"
 base64 = "0.13"
--- a/runtime/plaid/resources/plaid.toml
+++ b/runtime/plaid/resources/plaid.toml
@@ -13,7 +13,7 @@ r = ["test_shared_db_rule_1.wasm"]
 rw = ["test_shared_db_rule_2.wasm"]
 
 [storage.db]
-path = "/tmp/sled"
+sled_path = "/tmp/sled"
 
 # [storage.db]
 # table_name = "test-plaid"
--- a/runtime/plaid/resources/plaid.toml
+++ b/runtime/plaid/resources/plaid.toml
@@ -12,8 +12,17 @@ size_limit = { Limited = 50 }
 r = ["test_shared_db_rule_1.wasm"]
 rw = ["test_shared_db_rule_2.wasm"]
 
-[storage.sled]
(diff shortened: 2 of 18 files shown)


runtime/plaid/src/storage/dynamodb/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+        &self,
+        namespace: String,
+        key: String,
+        value: Vec<u8>,
+    ) -> Result<Option<Vec<u8>>, StorageError> {
+        let response = self
+            .client
+            .put_item()
+            .table_name(&self.table_name)
+            .item("namespace", AttributeValue::S(namespace))
```
michelemin, line 53: TODO for me: use constants

michelemin:
@obelisk After our investigation on the `sleep` issue, it's now clear that
1. My code was not using strongly consistent reads --> fixed ✅ .
2. If we have no `sleep` between requests and we are using multiple execution threads, requests can be executed in an overlapping fashion, leading to issues.
3. If we use only 1 execution thread and no `sleep`, then we send requests too fast. Plaid will reply before the processing is finished, which will unblock the following `curl`, and then another, and another until the end. Then we `sleep` for 2 seconds at the end of the test and then tear down everything. These 2 seconds are not enough to process the backlog, and we end up with failures.

Given point 3 above, there are two possibilities:
1. Keep the `sleep 0.5` in between requests
2. Sleep for a much longer time at the end of the test (something like 15 or 20 seconds)

I chose option 1 because it's the only one that works with multiple execution threads, and we don't want to be forced to use 1 execution threads in our tests because of this.

runtime/plaid/Cargo.toml
```diff
@@ -1,19 +1,19 @@
...
-version = "0.19.0"
+version = "0.20.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
 default = ["aws"]
-aws = ["aws-sdk-kms", "aws-config"]
+aws = ["aws-sdk-kms", "aws-sdk-dynamodb"]
```
michelemin, line 10: This is because we use `aws-config` in the secrets manager script, so this dependency cannot be made optional at the moment.

runtime/plaid/src/storage/mod.rs
```diff
@@ -115,9 +120,28 @@ pub trait StorageProvider {
...
 impl Storage {
     pub async fn new(config: Config) -> Result<Self, StorageError> {
-        let database = match config.sled {
-            Some(sled) => Box::new(sled::Sled::new(sled)?),
-            _ => return Err(StorageError::NoStorageConfigured),
+        // Try building a database by checking the config in this order:
+        // 1. sled
+        // 2. dynamodb (if the `aws` feature is enabled)
+        // If none of these works, then error because no storage is configured
+        #[cfg(feature = "aws")]
```
obelisk, line 127: This going to become unruly when we add other database backends (like Redis). Maybe use an enum with conditional types similar to what Rustica does?

obelisk (changes requested):

runtime/plaid/src/storage/dynamodb/mod.rs
```diff
@@ -0,0 +1,297 @@
...
+/// `table_name` - the name of the DynamoDB table  
+/// `namespace` - the namespace we want to query  
+/// `prefix` - [Optional] a prefix that keys must have in order to be returned by this query  
+/// `attributes_to_get` - A comma-separated list of attributes to get from DynamoDB
+async fn dynamodb_query(
+    client: &Client,
+    table_name: &str,
+    namespace: &str,
+    prefix: Option<&str>,
+    attributes_to_get: &str,
```
obelisk, line 200: I think this should be a `Vec` then turned into a str inside this function. I think that provides a cleaner, and clearer API

runtime/plaid/src/storage/dynamodb/mod.rs
```diff
@@ -0,0 +1,297 @@
...
+
+        Ok(everything)
+    }
+
+    async fn get_namespace_byte_size(&self, namespace: &str) -> Result<u64, StorageError> {
+        let all = self.fetch_all(namespace, None).await?;
+        let mut counter = 0u64;
+        for item in all {
+            // Count bytes for keys and values
+            counter = counter + item.0.as_bytes().len() as u64 + item.1.len() as u64;
```
obelisk, line 181: Nit. `+=`?

runtime/plaid/src/storage/mod.rs
michelemin: Right, good callout

runtime/plaid/src/storage/dynamodb/mod.rs
```diff
@@ -0,0 +1,297 @@
...
+    // Start filling in the mappings for names and attributes (to go back from aliased to real names)
+    let mut expression_attribute_names = vec![];
+    for (index, attribute) in attributes.iter().enumerate() {
+        expression_attribute_names.push((aliased[index].to_string(), attribute.to_string()));
+    }
+    expression_attribute_names.push(("#pk".to_string(), NAMESPACE.to_string()));
+
+    let mut expression_attribute_values = vec![];
+    expression_attribute_values.push((
+        ":pk_val".to_string(),
```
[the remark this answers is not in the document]
michelemin: See other comment

runtime/plaid/resources/plaid.toml
```diff
@@ -12,9 +12,18 @@ size_limit = { Limited = 50 }
...
 
+# [storage.db]
+# table_name = "test-plaid"
+
+# [storage.db.authentication]
+# access_key_id = "value here"
+# secret_access_key = "value here"
+# session_token = "value here"
+# region = "value here"
+
```
michelemin, line 26: This is a breaking change for the config.

obelisk (approved):

PR #94, opened 2025-04-02
Implement module signing

wbssbw:
The PR implements a mechanism to require that rules be signed prior to loading them. In `plaid.toml` this looks like:
```toml
[loading.rule_signing]
authorized_signers = [""]
signature_namespace = "SomeNamespace"
signature_dir = "/path/to/signatures"
```
When `[loading.rule_signing]` is present in `plaid.toml`, we require that ALL rules are signed by an authorized signer. This all or nothing approach greatly simplifies rule signing. When rule signing is enabled, the loading system:
1. Fetches the rule's signature from `signature_dir`
2. Validates the signature and checks that it was produced by one of `authorized_signers`
3. Computes the hash of the rule
4. Validates the signature against the rule's hash

This PR is as much a process change in Plaid administration as it is a code change. In practice, Plaid administrators will need to sign every rule prior to deployment. Depending on the environment, this can be scripted to reduce manual effort on the admin and CI checks can be put in place to ensure that every rule is signed prior to merge.

### DIFF
--- a/runtime/plaid/Cargo.toml
+++ b/runtime/plaid/Cargo.toml
@@ -49,7 +49,6 @@ reqwest_cookie_store = "0.8.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1"
 sled = { version = "0.34.7", optional = true }
-sha2 = "0.10.8"
 sshcerts = { version = "0.13.2", default-features = false }
 tar = "0.4.41"
 time = "0.3"
--- a/runtime/plaid/Cargo.toml
+++ b/runtime/plaid/Cargo.toml
@@ -49,6 +49,8 @@ reqwest_cookie_store = "0.8.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1"
 sled = { version = "0.34.7", optional = true }
+sha2 = "0.10.8"
+sshcerts = { version = "0.13.2", default-features = false }
 tar = "0.4.41"
 time = "0.3"
 tokio = { version = "1", features = ["full"] }
--- a/runtime/plaid/resources/plaid.toml
+++ b/runtime/plaid/resources/plaid.toml
@@ -58,6 +58,7 @@ test_mode_exemptions = [
 # [loading.module_signing]
 # authorized_signers = [""]
 # signature_namespace = "SomeNamespace"
+# signatures_required = 1
 
 [loading.persistent_response_size]
 "test_persistent_response.wasm" = 1024
--- a/runtime/plaid/resources/plaid.toml
+++ b/runtime/plaid/resources/plaid.toml
@@ -55,13 +55,11 @@ test_mode_exemptions = [
 
 
 # Uncomment this to require that all rules be signed by an authorized signer
-# [loading.module_signing]
-# authorized_signers = [
-#     "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIHkbe7gwx7s0dlApEEzpUyOAPrzPLy4czEZw/sh8m8rd me@home",
-# ]
-# signatures_required = 1
-# signature_namespace = "SomeNamespace" # Defaults to PlaidRule if not provided
-# signature_dir = "/path/to/signatures" # Defaults to ../module_signatures if not provided
+[loading.module_signing]
+authorized_signers = [
+    "{plaid-secret{public-key}}",
+]
+signatures_required = 1
 
 [loading.persistent_response_size]
 "test_persistent_response.wasm" = 1024
--- a/runtime/plaid/resources/plaid.toml
+++ b/runtime/plaid/resources/plaid.toml
@@ -56,7 +56,9 @@ test_mode_exemptions = [
 
 # Uncomment this to require that all rules be signed by an authorized signer
 # [loading.module_signing]
-# authorized_signers = [""]
+# authorized_signers = [
+#     "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIHkbe7gwx7s0dlApEEzpUyOAPrzPLy4czEZw/sh8m8rd me@home",
+# ]
 # signature_namespace = "SomeNamespace"
 # signatures_required = 1
 
--- a/runtime/plaid/resources/plaid.toml
+++ b/runtime/plaid/resources/plaid.toml
@@ -55,7 +55,7 @@ test_mode_exemptions = [
 
 
 # Uncomment this to require that all rules be signed by an authorized signer
-# [loading.rule_signing]
+# [loading.module_signing]
 # authorized_signers = [""]
 # signature_namespace = "SomeNamespace"
 
--- a/runtime/plaid/resources/plaid.toml
+++ b/runtime/plaid/resources/plaid.toml
@@ -53,6 +53,12 @@ test_mode_exemptions = [
     "test_shared_db_rule_2.wasm",
 ]
 
+
+# Uncomment this to require that all rules be signed by an authorized signer
+# [loading.rule_signing]
+# authorized_signers = [""]
+# signature_namespace = "SomeNamespace"
+
 [loading.persistent_response_size]
 "test_persistent_response.wasm" = 1024
 "test_sshcerts_usage.wasm" = 1024
--- a/runtime/plaid/resources/plaid.toml
+++ b/runtime/plaid/resources/plaid.toml
@@ -59,8 +59,9 @@ test_mode_exemptions = [
 # authorized_signers = [
 #     "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIHkbe7gwx7s0dlApEEzpUyOAPrzPLy4czEZw/sh8m8rd me@home",
 # ]
-# signature_namespace = "SomeNamespace"
 # signatures_required = 1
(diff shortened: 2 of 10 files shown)


runtime/plaid/Cargo.toml
```diff
@@ -49,6 +49,8 @@ reqwest_cookie_store = "0.8.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1"
 sled = { version = "0.34.7", optional = true }
+sha2 = "0.10.8"
+sshcerts = { version = "0.13.2", default-features = false }
```
obelisk, line 53: It's beautiful

runtime/plaid/Cargo.toml
```diff
@@ -49,6 +49,8 @@ reqwest_cookie_store = "0.8.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1"
 sled = { version = "0.34.7", optional = true }
+sha2 = "0.10.8"
```
obelisk, line 52: SHA2 is included in ring. Please remove this.

runtime/plaid/src/loader/mod.rs
```diff
@@ -128,6 +130,32 @@ pub struct Configuration {
     /// even if they have side effects.
     #[serde(default)]
     pub test_mode_exemptions: Vec<String>,
+    /// Configuration for module signing. If defined, we require that ALL
+    /// module are signed by an authorized signer
```
obelisk, line 134: By a set of authorized signers

runtime/plaid/src/loader/mod.rs
```diff
@@ -128,6 +130,32 @@ pub struct Configuration {
...
+}
+
+/// This structure defines the parameters required to validate signatures for modules.
+#[derive(Deserialize)]
+pub struct ModuleSigningConfiguration {
+    /// A list of authorized signer key fingerprints.
+    ///
+    /// This list should contain the fingerprints of the keys belonging to the authorized signers,
+    /// typically the administrators responsible for managing the Plaid instance.
+    pub authorized_signers: Vec<String>,
```
obelisk, line 145: Use a deserialization function to go right to `PublicKey`

runtime/plaid/src/loader/mod.rs
```diff
@@ -344,6 +372,50 @@ pub async fn load(
             }
         };
 
+        // Fetch and verify the corresponding signature over this module if we require
+        // rule signing
+        //
+        // If ANY rule does not have a valid signature, we panic instead of continuing because this should be
```
obelisk, line 378: Hmm, I'm not sure I agree with this because it means a problem with any rule will prevent Plaid from starting. Is that what we want? Erroring that a particular rule is not sufficiently/correctly signed seems a more sensible approach.

runtime/plaid/src/loader/mod.rs
obelisk: HashSet is also a better data structure here

runtime/plaid/src/loader/mod.rs
```diff
@@ -428,6 +500,68 @@ pub async fn load(
...
+    }
+
+    // Verify the signature using the module hash.
+    VerifiedSshSignature::from_ssh_signature(
+        module_hash,
+        ssh_signature,
+        &signing.signature_namespace,
+        Some(pubkey),
+    )
+    .map_err(|e| Errors::SshCertsError(e))?;
```
obelisk, line 560: Would call this SigningError. Even though we're using the `sshcerts` library, I don't think that context is important at this level

obelisk (changes requested):

runtime/plaid/src/loader/mod.rs
```diff
@@ -128,6 +132,52 @@ pub struct Configuration {
...
+
+/// This structure defines the parameters required to validate signatures for modules.
+#[derive(Deserialize)]
+pub struct ModuleSigningConfiguration {
+    /// A list of authorized signer key fingerprints.
+    ///
+    /// This list should contain the fingerprints of the keys belonging to the authorized signers,
+    /// typically the administrators responsible for managing the Plaid instance.
+    #[serde(deserialize_with = "pubkey_deserializer")]
+    pub authorized_signers: Vec<PublicKey>,
```
obelisk, line 148: Should be `HashSet`

runtime/plaid/src/loader/mod.rs
```diff
@@ -128,6 +132,52 @@ pub struct Configuration {
...
+    #[serde(default = "default_sig_dir")]
+    pub signatures_dir: String,
+    /// The namespace of the signature
+    pub signature_namespace: String,
+    /// The number of valid signatures required on each module
+    pub signatures_required: usize,
+}
+
+/// Deserializer for a public key
+fn pubkey_deserializer<'de, D>(deserializer: D) -> Result<Vec<PublicKey>, D::Error>
```
obelisk, line 160: Do we also want to enforce or at least warn if one of the keys doesn't have a comment?

runtime/plaid/src/loader/mod.rs
obelisk: This needs updating

runtime/plaid/src/loader/mod.rs
```diff
@@ -344,6 +394,56 @@ pub async fn load(
...
+            }
+
+            if valid_signatures.len() < signing.signatures_required {
+                error!(
+                    "Not enough valid signatures provided for {filename}. Got {} but needed {}",
+                    valid_signatures.len(),
+                    signing.signatures_required
+                );
+                continue;
+            }
```
obelisk, line 444: This should be a function

obelisk (changes requested):

runtime/plaid/src/loader/mod.rs
obelisk: @wbssbw points out `PublicKey` is not `Hash`

runtime/plaid/src/loader/mod.rs
wbssbw: We can warn but right now have no use for the comment so it might be unnecessary

runtime/plaid/src/loader/mod.rs
wbssbw: Fixed in [64a9225](https://github.com/obelisk/plaid/pull/94/commits/64a922523110b1504cacbda0ac7936763153c8ca)

obelisk (approved):

runtime/plaid/resources/plaid.toml
```diff
@@ -53,6 +53,15 @@ test_mode_exemptions = [
     "test_shared_db_rule_2.wasm",
 ]
 
+
+# Uncomment this to require that all rules be signed by an authorized signer
+# [loading.module_signing]
+# authorized_signers = [
+#     "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIHkbe7gwx7s0dlApEEzpUyOAPrzPLy4czEZw/sh8m8rd me@home",
+# ]
+# signature_namespace = "SomeNamespace"
```
obelisk, line 62: I would add a default here. "PlaidRule"

runtime/plaid/src/loader/signing.rs
```diff
@@ -0,0 +1,125 @@
...
+            .filter_map(|entry| {
+                entry.inspect_err(|e| {
+                    warn!("Bad entry in signature directory for {filename} - skipping. Error: {e}")
+                }).ok()
+            })
+            .collect::<Vec<_>>();
+
+    // If the number of available signature files does not exceed the required count,
+    // return an error immediately to avoid unnecessary processing.
+    if module_signatures.len() <= signing.signatures_required {
```
obelisk, line 36: Less than

wbssbw: Great catch

runtime/plaid/src/loader/signing.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/// Reads a signature file from the given directory entry.
+///
+/// Verifies the file ends with ".sig" and returns its content as a String,
+/// or an error if the file name is invalid or the file cannot be read.
+fn read_in_signature(signature_path: DirEntry) -> Result<String, Errors> {
+    let sig_filename = signature_path.file_name().to_string_lossy().to_string();
+    if !sig_filename.ends_with(".sig") {
+        return Err(Errors::BadFilename(".sig".to_string()));
```
obelisk, line 120: I think this error needs work.

wbssbw: Done in [7937ba5](https://github.com/obelisk/plaid/pull/94/commits/7937ba56438176819424e5fe155c18ce347d3b62)

obelisk (approved):

runtime/plaid/resources/plaid.toml
```diff
@@ -53,6 +53,14 @@ test_mode_exemptions = [
     "test_shared_db_rule_2.wasm",
 ]
 
+
+# Uncomment this to require that all rules be signed by an authorized signer
+[loading.module_signing]
+authorized_signers = [
+    "{plaid-secret{public-key}}",
```
michelemin, line 60: Does this have to be a secret?

PR #99, opened 2025-04-04
Add Message::id and change logbacks' persistence

### LINKED ISSUE #12: Logbacks from prior to LogSource Rewrite Are Stuck
While debugging an unrelated issue I noticed there are a bunch of warnings on start up from previous logbacks which are stuck because they were serialized before the `LogSource` rewrite.

The fix is pretty straight forward as all logs in the LogBack system are messages from other rules so we can convert them with no issue, we just need need a tool to do that, or to write support into Plaid.

I'm in favour of building a port tool.

michelemin:
Main changes:
* Every `Message` now has an ID. This is immediately useful for logbacks, and can be useful for other messages too in the future.
* When persisting logbacks in the DB, we use the `Message`'s ID as key, while the value is the serialization of the corresponding `DelayedMessage`.
* When deserializing and deleting logbacks from the DB, we only deserialize the new format. If data exists in the DB that was serialized in the old way(s), then it should be migrated with the tool that this PR introduces.

Note - In the previous format, the key was the JSON serialization of the message, while the value was the byte representation of the logback's delay.

Fixes #12

### DIFF
--- a/runtime/plaid/Cargo.toml
+++ b/runtime/plaid/Cargo.toml
@@ -61,6 +61,7 @@ tokio-tungstenite = { version = "0.23.1", features = [
 ] }
 toml = "0.5"
 totp-rs = "5.6.0"
+uuid = { version = "1", features = ["v4"] }
 url = "2.5.2"
 urlencoding = "2.1.3"
 warp = { version = "0.3", features = ["tls"] }
--- a/runtime/plaid/Cargo.toml
+++ b/runtime/plaid/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "plaid"
-version = "0.22.1"
+version = "0.22.2"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
@@ -91,4 +91,9 @@ path = "src/bin/secrets_manager.rs"
 [[bin]]
 name = "sled_ddb_migrator"
 path = "src/bin/sled_ddb_migrator.rs"
+required-features = ["sled", "aws"]
+
+[[bin]]
+name = "db_migrator"
+path = "src/bin/db_migrator.rs"
 required-features = ["sled", "aws"]
--- a/runtime/plaid/src/bin/db_migrator.rs
+++ b/runtime/plaid/src/bin/db_migrator.rs
@@ -2,10 +2,14 @@ use clap::{Arg, Command};
 use plaid::{data::DelayedMessage, executor::Message, storage::StorageProvider};
 use serde_json::Value;
 
-/// This function takes additional parameters and produces the function that defines the data migration.
-/// We use this pattern to be able to take arbitrary parameters that would not fit in the migration
-/// function's signature.
-fn create_migration_function(
+// These functions take additional parameters and produce the function that defines the data migration.
+// We use this pattern to be able to take arbitrary parameters that would not fit in the migration
+// function's signature.
+
+/// Create a migration function that harmonizes logbacks in the DB so that they comply with the latest
+/// serialization format. This ensures that there are no stuck logbacks in the DB that cannot be
+/// deserialized, executed, and then properly deleted.
+fn create_migration_function_harmonize_logbacks_0_22_2(
     log_source_if_missing: String,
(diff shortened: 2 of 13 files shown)


runtime/plaid/src/data/internal/mod.rs
```diff
@@ -65,9 +65,6 @@ pub struct Internal {
     storage: Option<Arc<Storage>>,
 }
 
-#[derive(Deserialize, Serialize)]
-struct InternalLog {}
-
```
michelemin, line 70: This was not used anywhere. Interestingly enough, it was not marked as dead code because of the `derive`. Didn't know.

runtime/plaid/Cargo.toml
```diff
@@ -59,6 +59,7 @@ tokio-tungstenite = { version = "0.23.1", features = [
 ] }
 toml = "0.5"
 totp-rs = "5.6.0"
+uuid = { version = "1", features = ["v4"] }
```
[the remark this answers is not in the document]
obelisk: I think this is fine

runtime/plaid/src/executor/mod.rs
```diff
@@ -35,14 +35,16 @@ pub struct ResponseMessage {
...
     /// The message channel that this message is going to run on
     pub type_: String,
     /// The data passed to the module
     pub data: Vec<u8>,
     /// Any headers the module will have access to, while processing this message
-    pub headers: Option<HashMap<String, Vec<u8>>>,
+    pub headers: HashMap<String, Vec<u8>>,
     /// Any query parameters the module will have access to, while processing this message
-    pub query_params: Option<HashMap<String, Vec<u8>>>,
+    pub query_params: HashMap<String, Vec<u8>>,
```
michelemin, line 47: These are not Option anymore because we mandate one and only one DB format. If the DB is not in that format, the migrator tool can be used.

runtime/plaid/src/bin/db_migrator.rs
```diff
@@ -0,0 +1,178 @@
+use clap::{Arg, Command};
+use plaid::{data::DelayedMessage, executor::Message, storage::StorageProvider};
+use serde_json::Value;
+
+/// This function takes additional parameters and produces the function that defines the data migration.
+/// We use this pattern to be able to take arbitrary parameters that would not fit in the migration
+/// function's signature.
+fn create_migration_function(
```
michelemin, line 8: Note - `unwraps` are used because nothing should ever fail here. If it does, a migration can be repeated basically without risks, since operations are idempotent.

obelisk, line 8: Should we name this with a version, at least what it migrates to? I think in the future that would help be clear about what functions need to be run to get the database into a fully migrated state

runtime/plaid/src/bin/db_migrator.rs
```diff
@@ -0,0 +1,178 @@
...
+                        "source".to_string(),
+                        serde_json::to_value(plaid_stl::messages::LogSource::Logback(
+                            log_source_if_missing.clone(),
+                        ))
+                        .unwrap(),
+                    );
+                }
+                if v.get("accessory_data").is_some() {
+                    // `accessory_data` is present: we remove it and add `headers` and `query_params`.
+                    // Note: the content of `accessory_data` is lost.
```
obelisk, line 47: We don't expect this to be a problem correct?

michelemin: Technically, it could be a problem if there is a stuck message that must be executed and it relies on some accessory_data that was serialized at the time of inserting into the DB.
Practically speaking, the problem would be understanding if what is inside accessory_data is a secret, a header, a query param, and then putting that into the correct "bucket". Also, things have since changed and secrets are no longer associated to messages but to modules. In short, I think we should drop accessory_data and accept the migration is not bullet-proof.

runtime/plaid/src/bin/db_migrator.rs
michelemin: Great call out, yes, it will make it much clearer. Updated in https://github.com/obelisk/plaid/pull/99/commits/a81bd9c78369fca7bede31bbadec5fb0acd7b014

obelisk (approved):

PR #105, opened 2025-04-08
Add `show_log_on_error` configuration flag to omit logs from logs

wbssbw:
In some instances, the log sent to a rule is sensitive and should not be logged to external logging sources on rule execution failure. To combat this, we introduce an optional `show_log_on_error` (defaults to `true`) flag in the `LoggingConfiguration` struct. When set to `false`, the logging subsystem replaces the log sent to the rule with an empty string.

Currently, this is a global setting in the runtime and applies to all rules but could be applied at a more granular level by making `show_log_on_error` into a list of rule names


obelisk (commented):
I think I would want the logic flipped on this. `show_log_on_error` perhaps? I think that is more clear because we're not doing actual redaction.

PR #111, opened 2025-04-10
Add Cloudtrail data generator

michelemin:
Following the same pattern as the other data generators.

This might be obsolete because of #119

### DIFF
--- a/runtime/plaid/src/data/cloudtrail/mod.rs
+++ b/runtime/plaid/src/data/cloudtrail/mod.rs
@@ -1,0 +1,178 @@
+use std::num::NonZeroUsize;
+
+use aws_sdk_cloudtrail::Client;
+use aws_sdk_kms::primitives::DateTime;
+use crossbeam_channel::Sender;
+use lru::LruCache;
+use plaid_stl::messages::{Generator, LogSource, LogbacksAllowed};
+use serde::Deserialize;
+use time::OffsetDateTime;
+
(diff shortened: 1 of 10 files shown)


runtime/plaid/src/data/cloudtrail/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+/// This function provides the default sleep duration in milliseconds.
+/// It is used as the default value for deserialization of the `sleep_duration` field,
+/// of `CloudtrailConfig` in the event that no value is provided.
+fn default_sleep_milliseconds() -> u64 {
+    1000
+}
+
+fn default_canon_time() -> u64 {
```
obelisk, line 38: Citation needed

runtime/plaid/src/data/cloudtrail/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+        }
+    }
+}
+
+impl DataGenerator for &mut Cloudtrail {
+    async fn fetch_logs(
+        &self,
+        since: time::OffsetDateTime,
+        until: time::OffsetDateTime,
+    ) -> Result<Vec<super::DataGeneratorLog>, ()> {
```
obelisk, line 80: Import this type so we're not specifying each time

obelisk (changes requested):

michelemin:
Closing as obsolete. Let's focus on #119 instead.

PR #115, opened 2025-05-02
Slack API Refactor

obelisk:
I revisited the Slack API because I wanted to add the ability to fetch user info (specifically timezone data) and I was unhappy with how I previously structured the code. There was no clear separation of responsibilities and it was hard to see how APIs were implemented and especially difficult to add new ones. 

This changes all of that and couples the STL and the Plaid library together to throw compile time errors for inconsistencies rather than runtime `ApiError::MissingParameter` errors.


timweri (approved):
Clean!

PR #118, opened 2025-05-03
Split config

obelisk:
Split the plaid.toml config files into multiple TOML files which are read from a given directory.
These files are then concatenated back into a larger file that is parsed and consumed by the runtime, but having separate files is easier to manage. Also, this approach is backwards compatible with having a single plaid.toml file.

Files are concatenated in alphabetical order.

Note: It's important that everything is under a top-level TOML key. For this reason, we refactor the config so that execution_threads is now under the [executor] key. Otherwise, loose items might end up in a table they don't belong to.

TODO
- [x] Skip Slack tests if secrets are not available


obelisk:
Let's just make sure the new Slack tests pass before pulling this over to your PR

PR #121, opened 2025-05-05
Dedicated threads per log type

michelemin:


runtime/plaid/src/lib.rs
```diff
@@ -16,6 +20,50 @@ pub mod logging;
...
+    /// Thread pools dedicated to specific log types.
+    /// Mapping { log_type --> thread_pool }
+    pub dedicated_pools: HashMap<String, ThreadPool>,
+}
+
+impl ExecutionThreadPools {
+    /// Create a new ExecutionThreadPools object by initializing only the thread
+    /// pool for general processing. Other thread pools, if present, must be
+    /// added separately by inserting into the `dedicated_pools` map.
+    pub fn new(num_threads: u8, queue_size: usize) -> Self {
```
wbssbw, line 59: I think you could have this take a `&ExecutorConfig` and set all the pools up here, rather than inserting the dedicated pools later on

michelemin: Good call out: done!

runtime/plaid/src/bin/plaid.rs
```diff
@@ -93,38 +96,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let modules = Arc::new(loader::load(config.loading, storage.clone()).await.unwrap());
     let modules_by_name = Arc::new(modules.get_modules());
 
-    info!(
-        "Starting the execution threads of which {} were requested",
-        config.executor.execution_threads
-    );
+    // Print information about the threads we are starting
+    info!("Starting {} execution threads for general execution", exec_thread_pools.general_pool.num_threads);
+    if !exec_thread_pools.dedicated_pools.is_empty() {
```
obelisk, line 101: I would just do a iteration here, no need for the "Additionally" message :)

runtime/plaid/src/bin/plaid.rs
```diff
@@ -154,12 +159,27 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                            }
+                            None => {
+                                // We have no dedicated thread pool for this type: just send to the general one.
+                                if let Err(e) = exec_thread_pools.general_pool.sender.try_send(message) {
+                                    match e {
+                                        TrySendError::Full(_) => error!("Queue Full! [{}] log dropped!", webhook_configuration.log_type),
+                                        // TODO: Have this actually cause Plaid to exit
+                                        TrySendError::Disconnected(_) => panic!("The execution system is no longer accepting messages. Nothing can continue."),
+                                    }
+                                }
```
obelisk, line 182: Should we move this into an impl on the executor object? That would let us hide this away from the main.rs which we need to restructure anyway

runtime/plaid/src/config.rs
```diff
@@ -110,6 +110,9 @@ pub struct ExecutorConfig {
     /// The maximum number of logs in the queue to be processed at once
     #[serde(default = "default_log_queue_size")]
     pub log_queue_size: usize,
+    /// Number of threads dedicated to specific log types.
+    /// This is a mapping {log type --> num threads}.
+    pub dedicated_threads: Option<HashMap<String, u8>>,
```
obelisk, line 115: I wouldn't have this be an option. I would have it just be a `HashMap` and then give it a default for deserialization

runtime/plaid/src/config.rs
```diff
@@ -110,6 +110,9 @@ pub struct ExecutorConfig {
     /// The maximum number of logs in the queue to be processed at once
     #[serde(default = "default_log_queue_size")]
     pub log_queue_size: usize,
```
obelisk, line 112: Do we want to allow this to be set for dedicated_threads as well? Or does that give too much flexibility

obelisk (changes requested):

runtime/plaid/src/config.rs
michelemin: Yeah I was considering that but thought that one value should be sufficient.
The config would not get much more complex, because we could still have default values... We could say

```
[executor.dedicated_threads."log_type_1"]
num_threads = 3
queue_size = 4096

 [executor.dedicated_threads."log_type_2"]
num_threads = 2
```

@obelisk What do you think?

runtime/plaid/src/bin/plaid.rs
michelemin: Yeah, I got carried away 😄

runtime/plaid/src/bin/plaid.rs
michelemin: Yeah it's cleaner this way: done!

runtime/plaid/src/config.rs
michelemin: OK, then just `#[serde(default)]` should work because `HashMap` implements `Default`, which gives an empty map.

runtime/plaid/src/config.rs
michelemin: @obelisk It's implemented in a separate commit, so it's easy to roll it back. But, now that I see it written out, I think we should keep it :)

runtime/plaid/src/bin/plaid.rs
```diff
@@ -93,38 +96,38 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        exec_thread_pools.clone(),
         modules.get_channels(),
         api,
         storage,
-        config.executor.execution_threads,
         els.clone(),
         performance_sender.clone()
     );
 
     let _executor = Arc::new(executor);
```
obelisk, line 117: This should be renamed as we're now using it (remove the leading underscore)

runtime/plaid/src/bin/plaid.rs
```diff
@@ -93,38 +96,38 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     let modules_by_name = Arc::new(modules.get_modules());
 
-    info!(
-        "Starting the execution threads of which {} were requested",
-        config.executor.execution_threads
-    );
+    // Print information about the threads we are starting
+    info!("Starting {} execution threads for general execution. Log queue size = {}", exec_thread_pools.general_pool.num_threads, exec_thread_pools.general_pool.sender.capacity().unwrap_or_default());
+    for (log_type, tp) in &exec_thread_pools.dedicated_pools {
+        let thread_or_threads = if tp.num_threads == 1 {"thread"} else {"threads"};
```
obelisk, line 102: Dedication.

runtime/plaid/src/executor/mod.rs
```diff
@@ -409,24 +413,6 @@ fn process_message_with_module(
     els: Logger,
     performance_mode: Option<Sender<ModulePerformanceMetadata>>,
 ) -> Result<(), ExecutorError> {
-    // For every module that operates on that log type
-    // Mark this rule as currently being processed by locking the mutex
-    // This lock will be dropped at the end of the iteration so we don't
-    // need to handle unlocking it
-    let _lock = match module.concurrency_unsafe {
```
obelisk, line 416: Did you check all references to this were removed from the configuration as well?

runtime/plaid/src/executor/mod.rs
```diff
@@ -628,6 +615,44 @@ impl Executor {
...
+        Self {
+            _handles,
+            thread_pools,
+        }
+    }
+
+    /// Execute a message coming from a webhook, by sending it to the appropriate thread pool.
+    /// That will be the thread pool dedicated to the message's type, if one exists, or the
+    /// default thread pool for general execution.
+    pub fn execute_webhook_message(
```
obelisk, line 642: Nice

obelisk (approved):

PR #123, opened 2025-05-06
Support pagination in data generators

michelemin:
Add support for pagination to `GitHub` and `Okta` data generators. This is done by leveraging the `link` header, which is included by both APIs in a multi-page response.

With this change, we mandate that `fetch_logs` return all logs that happened in the `since..until` time span. If this requires handling pagination, this must be done inside this function.

Note: we also remove the mechanism known as "kicking the window", where we were manually changing `last_seen` to ensure the data generator did not get stuck. The reason is that now we don't limit ourselves to the first page of logs: instead we fetch all the logs in the `since..until` time span. Therefore, as `until` naturally moves forward (since it's calculated as `now - canon_time`) we will automatically get newer logs together with older ones, which will be discarded thanks to the cache.

TODO
- [x] Make cache size configurable

### DIFF
--- a/modules/Cargo.lock
+++ b/modules/Cargo.lock
@@ -180,7 +180,7 @@ checksum = "57c0d7b74b563b49d38dae00a0c37d4d6de9b432382b2892f0574ddcae73fd0a"
 
 [[package]]
 name = "plaid_stl"
-version = "0.23.2"
+version = "0.23.3"
 dependencies = [
  "base64 0.13.1",
  "chrono",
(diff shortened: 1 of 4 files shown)


obelisk:
Are we waiting on the configurable cache size or should we review and merge this?

michelemin:
@obelisk Sorry, I had not ticked the box but the configurable cache size is already in. So for me we can review and merge.

obelisk:
Need to resolve merge conflicts

obelisk (approved):

PR #124, opened 2025-05-06
Add CRON schedule support to interval job generator

wbssbw:
The current implementation of the interval data generator is limited in the use cases it supports: jobs can only run every `N` seconds - no more, no less. This PR aims to give Plaid administrators more control over interval jobs by adding support for CRON. 

What we would previously define as:
```toml
[data.interval]
[data.interval.jobs]
[data.interval.jobs."every_minute"]
interval = 60
log_type = "some-log-type"
data = "Some data sent to rule"
```

we now define as:
```toml
[data.interval]
[data.interval.jobs]
[data.interval.jobs."every_minute"]
schedule = "0 * * * * * *"
log_type = "some-log-type"
data = "Some data sent to rule"
```

We can also define jobs like:
```toml
[data.interval]
[data.interval.jobs]
[data.interval.jobs."every_MWF_at_1200_UTC"]
schedule = "0 0 12 * * Mon,Wed,Fri *"
log_type = "some-log-type"
data = "Some data sent to rule"
```

The `schedule` syntax is very close to traditional CRON syntax, but not exactly the same. Rather than the usual five fields (`min hour dom mon dow`), the `cron` crate uses **seven** required fields in this order:

```
sec min hour day-of-month month day-of-week year
```

All seven fields support the familiar operators:

- `*` for "every …"
- comma-separated lists: `Mon,Wed,Fri`
- numeric or named ranges: `May–Aug`
- step values: `*/15` or `2022/2` in the year field

#### What’s different from "classic" cron?
1. **Seconds field**  
   You must specify a seconds slot (e.g. `0 …`).  
2. **Year field**  
   You must also specify a year (or range/step) at the very end.  
3. **No shortcuts**  
   There are _no_ `@daily`/`@hourly` aliases, everything is spelled out.  
4. **Names or numbers**  
   Months (`Jan` – `Dec`) and weekdays (`Mon` – `Sun`) accept their English names _or_ numeric values.


obelisk (dismissed):
Conflicts

PR #125, opened 2025-05-08
AWS DynamoDB API 

harry-anderson:
Adds an AWS DynamoDB API for put_item, delete_item and query operations against a table. The api is intended for modules to use to write/read from dynamodb, separate from the storage api. The API is json based, takes json as input and returns json output. JSON object is translated to a record in dynamoDB, each key is mapped to an 'attribute' i.e. column. Not all options have been wired up yet e.g. query.filter_condition().

To use this API effectively it is recommended to follow[ Dynamo Single Table Design](https://aws.amazon.com/blogs/compute/creating-a-single-table-design-with-amazon-dynamodb/) access patterns. e.g. store items with composite primary key made up of two attributes of the json object. the partition key e.g pk and the sort key (range key) e.g. timestamp. Then you can add as many extra metadata fields as you want to your items. When you want to query for items in your table, you can query based on partition key AND/OR sort (range) key, or query another index by its name.

Example
```rust
  // Example: PutItem with all attribute types
    let item_json = serde_json::json!({
        "pk": "124",
        "timestamp": "124",
        "name": "Jane Doe",
        "age": 33,
        "is_active": true,
        "null_field": null,
        "scores": [95, 88, 92], // List
        "metadata": { // Map
            "city": "New York",
            "country": "USA"
        },
        "tags": ["dev", "rust", "aws"], // String Set
        "ratings": [4.5, 3.8, 5.0], // Number Set
        "binaries": [ // Binary Set (base64-encoded)
            BASE64_STANDARD.encode("data1"),
            BASE64_STANDARD.encode("data2")
        ],
        "binary_field": { "_binary": BASE64_STANDARD.encode("binary_data") } // Binary
    });
    let item_hm = serde_json::from_value::<HashMap<String, Value>>(item_json).unwrap();
    let input = PutItemInput {
        table_name: String::from("access_events"),
        item: item_hm,
        return_values: Some(String::from("ALL_OLD")),
        ..Default::default()
    };
    let parms = serde_json::to_string(&input).unwrap();
    let x = client.put_item(&parms).await?;

    println!("put_item output {x:?}");

    let input = QueryInput {
        table_name: String::from("access_events"),
        key_condition_expression: String::from("#pk = :val"),
        expression_attribute_names: Some(HashMap::from([("#pk".to_string(), "pk".to_string())])),
        expression_attribute_values: Some(HashMap::from([(
            ":val".to_string(),
            Value::String(String::from("124")),
        )])),
        ..Default::default()
    };
    let params = serde_json::to_string(&input).unwrap();
    let x = client.query(&params).await.unwrap();

    println!("query output {}", serde_json::to_string_pretty(&x).unwrap());

    let input = DeleteItemInput {
        table_name: String::from("access_events"),
        key: HashMap::from([
            (String::from("pk"), Value::String(String::from("124"))),
            (
                String::from("timestamp"),
                Value::String(String::from("124")),
            ),
        ]),
        return_values: Some(String::from("ALL_OLD")),
        ..Default::default()
    };

    let params = serde_json::to_string(&input).unwrap();
    let x = client.delete(&params).await;

    println!("delete {:?}", x);


```
output
```
put_item output None
query output [
  {
    "age": 33,
    "binaries": [
      "ZGF0YTE=",
      "ZGF0YTI="
    ],
    "binary_field": {
      "_binary": "YmluYXJ5X2RhdGE="
    },
    "is_active": true,
    "metadata": {
      "city": "New York",
      "country": "USA"
    },
    "name": "Jane Doe",
    "null_field": null,
    "pk": "124",
    "ratings": [
      4.5,
      3.8,
      5
    ],
    "scores": [
      88,
      92,
      95
    ],
    "tags": [
      "aws",
      "dev",
      "rust"
    ],
    "timestamp": "124"
  }
]
delete Ok(Some(Object {"age": Number(33), "binaries": Array [String("ZGF0YTE="), String("ZGF0YTI=")], "binary_field": Object {"_binary": String("YmluYXJ5X2RhdGE=")}, "is_active": Bool(true), "metadata": Object {"city": String("New York"), "country": String("USA")}, "name": String("Jane Doe"), "null_field": Null, "pk": String("124"), "ratings": Array [Number(4.5), Number(3.8), Number(5)], "scores": Array [Number(88), Number(92), Number(95)], "tags": Array [String("aws"), String("dev"), String("rust")], "timestamp": String("124")}))

### DIFF
--- a/runtime/plaid-stl/src/aws/dynamodb.rs
+++ b/runtime/plaid-stl/src/aws/dynamodb.rs
@@ -57,7 +57,7 @@ pub fn put_item(input: PutItemInput) -> Result<PutItemOutput, PlaidFunctionError
         new_host_function_with_error_buffer!(aws_dynamodb, put_item);
     }
 
-    let input = serde_json::to_string(&input).unwrap();
+    let input = serde_json::to_string(&input).map_err(|_| PlaidFunctionError::InternalApiError)?;
 
     let mut return_buffer = vec![0; RETURN_BUFFER_SIZE];
 
@@ -88,7 +88,7 @@ pub fn delete_item(input: DeleteItemInput) -> Result<DeleteItemOutput, PlaidFunc
         new_host_function_with_error_buffer!(aws_dynamodb, delete_item);
     }
 
-    let input = serde_json::to_string(&input).unwrap();
+    let input = serde_json::to_string(&input).map_err(|_| PlaidFunctionError::InternalApiError)?;
 
     let mut return_buffer = vec![0; RETURN_BUFFER_SIZE];
 
@@ -119,7 +119,7 @@ pub fn query(input: QueryInput) -> Result<QueryOutput, PlaidFunctionError> {
         new_host_function_with_error_buffer!(aws_dynamodb, query);
     }
 
-    let input = serde_json::to_string(&input).unwrap();
+    let input = serde_json::to_string(&input).map_err(|_| PlaidFunctionError::InternalApiError)?;
 
     let mut return_buffer = vec![0; RETURN_BUFFER_SIZE];
 

### CODE REVIEW

File: runtime/plaid/src/apis/aws/dynamodb.rs [hunk 1]
```diff
@@ -0,0 +1,538 @@
...
+        condition_expression: Option<&str>,
+        return_values: Option<&str>,
+    ) -> Result<Option<Value>, String> {
+        if let Some(write_access) = self.write.get(table_name) {
+            // check if this module has write access to this table
+            if !write_access.contains(&module.to_string()) {
+                return Err(format!("{module} no write access for {table_name}"));
+            }
+        } else {
+            return Err(format!("{module} no write access for {table_name}"));
```
michelemin, line 73: Maybe we can return a different error, like "unknown table".

runtime/plaid/src/apis/aws/dynamodb.rs
```diff
@@ -0,0 +1,538 @@
...
+        table_name: &str,
+        index_name: Option<&str>,
+        key_condition_expression: &str,
+        expression_attribute_names_json: Option<Value>,
+        expression_attribute_values_json: Option<Value>,
+    ) -> Result<Vec<Value>, String> {
+        if let Some(read_access) = self.read.get(table_name) {
+            // check if this module has read access to this table
+            if !read_access.contains(&module.to_string()) {
+                return Err(format!("{module} no write access for {table_name}",));
```
michelemin, line 289:
```suggestion
                return Err(format!("{module} no read access for {table_name}",));
```

runtime/plaid/src/apis/aws/dynamodb.rs
```diff
@@ -0,0 +1,538 @@
...
+    /// - `ApiKey`: Uses explicit credentials, including an access key ID, secret access key, and region.
+    authentication: AwsAuthentication,
+    /// Configured writers - maps a table name to a list of rules that are allowed to write it
+    write: HashMap<String, HashSet<String>>,
+    /// Configured readers - maps a table name to a list of rules that are allowed to read it
+    read: HashMap<String, HashSet<String>>,
+}
+
+/// Represents the KMS API that handles all requests to KMS
+pub struct DynamoDb {
```
obelisk, line 29: Add a comment here that if Plaid is configured with the DynamoDB database backend, sharing tables here will lead to undefined behaviour.

runtime/plaid/src/apis/aws/dynamodb.rs
```diff
@@ -0,0 +1,538 @@
...
+    pub async fn put_item(
+        &self,
+        module: Arc<PlaidModule>,
+        table_name: &str,
+        item_json: Value,
+        expression_attribute_names_json: Option<Value>,
+        expression_attribute_values_json: Option<Value>,
+        condition_expression: Option<&str>,
+        return_values: Option<&str>,
+    ) -> Result<Option<Value>, String> {
```
obelisk, line 66: This error type will need to change when you start linking it with functions.rs

runtime/plaid/src/apis/aws/dynamodb.rs
```diff
@@ -12,31 +14,42 @@ use serde::Deserialize;
...
     /// The underlying KMS client used to interact with the KMS API.
+    /// NOTE: if Plaid is configured with the DynamoDB database backend, sharing tables here will lead to undefined behaviour
     client: Client,
     /// Configured writers - maps a table name to a list of rules that are allowed to write it
     write: HashMap<String, HashSet<String>>,
     /// Configured readers - maps a table name to a list of rules that are allowed to read it
     read: HashMap<String, HashSet<String>>,
 }
 
+// TODO move to STL
```
michelemin, line 41: TODO in code

runtime/plaid/src/apis/aws/dynamodb.rs
```diff
@@ -0,0 +1,472 @@
...
+                        return Err(ApiError::BadRequest);
+                    }
+                    Ok(())
+                } else if let Some(write_access) = self.write.get(table_name) {
+                    // check if this module has write access to this table
+                    if !write_access.contains(&module.to_string()) {
+                        return Err(ApiError::BadRequest);
+                    }
+                    Ok(())
+                } else {
```
michelemin, line 86: Say that we have
```
my_table.read = [rule1, rule2]
my table.write = [rule3]
```
and we are checking if `rule3` can read. The answer should be yes, because it's in write.

However, with this code, we would return an error after the first `if`.

runtime/plaid/src/apis/aws/dynamodb.rs
```diff
@@ -0,0 +1,472 @@
...
+        let client = Client::new(&sdk_config);
+
+        Self {
+            client,
+            write,
+            read,
+        }
+    }
+
+    fn allow_operation(
```
michelemin, line 66: I would perhaps rename, because this looks like we are performing the action of allowing, while in reality we are only checking.

runtime/plaid/src/apis/aws/dynamodb.rs
```diff
@@ -0,0 +1,472 @@
...
+            .set_item(item)
+            .set_expression_attribute_names(expression_attribute_names)
+            .set_expression_attribute_values(expression_attribute_values)
+            .set_condition_expression(condition_expression.map(|x| x.to_string()))
+            .set_return_values(return_values)
+            .send()
+            .await
+            .map_err(|e| ApiError::DynamoDbPutItemError(e))?;
+
+        let attributes = match output.attributes() {
```
michelemin, line 139: You can use `.map()` here

PR #126, opened 2025-05-12
Move to Wasmer 6

obelisk:
While dying (flu) it felt like it might be an easy lift to move to Wasmer 6 and get the compute benefits of that (especially with the LLVM backend https://wasmer.io/posts/announcing-wasmer-6-closer-to-native-speeds

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .DS_Store
 
 # Artefacts
+/modules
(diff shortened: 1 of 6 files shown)


obelisk:
This also removes a bunch of compression libraries (likely due to wasmer deciding they don't need zip with all features)

PR #128, opened 2025-05-14
Increase sleep duration when fetching Okta logs

michelemin:
On a successful call to Okta's `/logs` endpoint, i.e., a call that yields new logs, we now sleep for 1 second before making a new request.

The previous value (1 ms) was not "wrong" but it made it likely for this to happen:
* call `/logs` and get some new logs
* sleep for 1 ms
* call `/logs` and get the same logs (very likely because so little time has passed)
* exit the loop, which will cause the thread to sleep for 10 seconds

With this change, we increase the probability that new logs come in during the sleep time, thus pulling in logs at a more consistent pace.

@wbssbw Unless I am missing something and there was a reason to set that value so low.


wbssbw:
Yeah, I don't remember the reasoning behind defaulting to 1 millisecond if no value is provided. Moving to 1 second makes sense to me

obelisk (approved):

PR #131, opened 2025-05-16
Support request bodies and headers on `GET` routes

wbssbw:
By default `GET` webhooks drop any request body; this PR changes the `GET` handlers to read up to 256 KiB from the body and populate the `Message` struct with those bytes before sending into the execution system. In practice it may be unusual for a `GET` to accept a body, but this change unlocks additional use cases that would otherwise require modifying the implementation of `POST` routes to return data

In addition, we now collect all headers and pass in the requested ones from the webhook's configuration to the `Message` struct


michelemin:
> read up to 256 KiB from the body 

Are we enforcing this? I saw it in a previous commit, but cannot find it anymore.

wbssbw:
> > read up to 256 KiB from the body
> 
> @wbssbw Are we enforcing this? I saw it in a previous commit, but cannot find it anymore.

Good callout - we once were, but Warp's length check expected a `Content-Length` header which caused normal GET requests to fail. I've removed it for now but will add a similar check on the body length that doesn't expect a header

obelisk (approved):

PR #137, opened 2025-05-28
Implement S3 APIs

wbssbw:
This PR introduces a new S3 module that allows Plaid rules to read from and write to S3 buckets with configurable access controls. The module integrates with AWS SDK clients for S3 and supports both direct object access and presigned URL generation.

## Summary of Changes
- Implements the S3 struct that wraps an AWS S3 client and bucket configuration.
- Adds `put_object` API for uploading objects to S3, gated by Write permissions.
- Adds `get_object` API for downloading objects or generating presigned URLs, gated by Read permissions.
- The response from `get_object` is serialized as either raw object data or a presigned URL depending on the request mode. To prevent a rule from forcing the runtime into reading a large object into memory, we implement the following checks:
  - **STL**: Call `get_object_attributes` to fetch metadata (notably size) about the object and then try to allocate a buffer of that size. If the object’s size is greater than the memory allocated to the rule, this will cause an OOM error.
  - **Runtime**: Introduce an optional (defaults to 4 MiB) `max_object_fetch_size` in the S3 configuration. This gives a plaidmin full control over how objects should be fetched and also acts as an "opt-in" when needing to fetch large objects, effectively forcing the plaidmin to accept the risk.
- Access control is enforced by checking the configured rule associated with the calling module (`PlaidModule`) against the bucket's allowed rules.

### Configuration
The `S3Config` structure includes:
- `authentication`: how the AWS client should authenticate (IAM or explicit credentials).
- `max_object_fetch_size` (optional): the maximum object size that Plaid will read into memory
- `bucket_configuration`: a mapping of bucket IDs to a list of rules and associated permissions.

Example (using IAM):
```toml
[apis."aws"]
[apis."aws"."s3"]
max_object_fetch_size = 5242880 # 5 MiB
[apis."aws"."s3".authentication]
[apis."aws"."s3".bucket_configuration]
[apis."aws"."s3".bucket_configuration."example-bucket"]
r = ["rule1.wasm"]
rw = ["rule2.wasm"]
```


runtime/plaid/src/apis/aws/s3.rs
```diff
@@ -0,0 +1,206 @@
...
+        self.client
+            .put_object()
+            .bucket(request.bucket_id)
+            .body(ByteStream::from(request.object))
+            .key(request.object_key)
+            .send()
+            .await
+            .map_err(S3Errors::S3PutObjectError)?;
+
+        Ok(String::new())
```
michelemin, line 180: Why do we need this?

obelisk (changes requested):
Rebase required

PR #139, opened 2025-05-30
Add first load test rule

michelemin:
Add a rule to perform load testing on Plaid. Also, update the locustfile to use that rule.

### DIFF
--- a/modules/load_tests/load_test/src/lib.rs
+++ b/modules/load_tests/load_test/src/lib.rs
@@ -1,0 +1,57 @@
+mod cache;
+mod regex;
+
+use plaid_stl::{
+    entrypoint_with_source_and_response,
+    messages::LogSource,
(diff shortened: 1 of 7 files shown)


obelisk (approved):

obelisk:
I still think you should make those `nth` changes though

modules/load_tests/load_test/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+    input_for_regex: Option<String>,
+    regex_to_use: Option<String>,
+    get_time: Option<bool>,
+    get_random_bytes: Option<u16>,
+    use_cache: Option<bool>,
+    post_slack: Option<bool>,
+}
+
+fn main(log: String, source: LogSource) -> Result<Option<String>, i32> {
+    // Only accept GET
```
michelemin, line 24: @michelemin Fix this.

PR #149, opened 2025-06-04
When adding a collaborator, do not fail on 201 response

michelemin:
When adding a collaborator to a repo, we were treating 204 as the only success response. However, the API can also return 201, if an invitation has been sent.

Documentation: https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#add-a-repository-collaborator--status-codes


obelisk:
It's the end of an era

obelisk (approved):

PR #158, opened 2025-06-06
Do not use unwrap when pulling randomness for splay

michelemin:
When generating the timing oﬀset for tasks ran on a schedule it is possible that the source of the random oﬀset will
cause the program to panic. This is due to the use of `unwrap()` in generating this random oﬀset.

While cases in which `srand.fill()` returns an error are unlikely to occur in practice, this can happen with edge cases
such as a lack of system entropy or when the code is run on a custom system.

This PR removes the `unwrap()` and, in case `srand.fill()` fails, defaults to a safe value.

### DIFF
--- a/runtime/plaid/src/data/interval/mod.rs
+++ b/runtime/plaid/src/data/interval/mod.rs
@@ -153,8 +153,13 @@ impl Interval {
 
             // Generate a random number between 0 and splay
             let mut bytes = [0u8; 8];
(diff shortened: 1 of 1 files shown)


obelisk (changes requested):
Is this actually correct? My fear is that we would do this and the exact problem we're trying to prevent (all intervals running at the same time) would manifest.

PR #159, opened 2025-06-10
Persist data generators' state across reboots

michelemin:
When the client restarts, previously seen logs and associated since and until values are discarded, resulting in a
loss of continuity and a gap in log processing during downtime.

With this PR, we persist the value of `last_seen` and the LRU cache for the logs which we have already seen. These will enable Plaid to start from where we had left off in case of a reboot.

### DIFF
--- a/runtime/plaid/src/data/mod.rs
+++ b/runtime/plaid/src/data/mod.rs
@@ -23,6 +23,10 @@ use time::OffsetDateTime;
 
 pub use self::internal::DelayedMessage;
 
+const DATA_GENERATOR_STORAGE_PREFIX: &str = "__DATA_GENERATOR";
+const LAST_SEEN: &str = "last_seen";
+const ALREADY_SEEN_UUIDS: &str = "already_seen_uuids";
+
 // Configure data sources that Plaid will use fetch data itself and
 // send to modules
 #[derive(Deserialize)]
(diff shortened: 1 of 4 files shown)


runtime/plaid/src/data/mod.rs
```diff
@@ -230,15 +258,152 @@ fn get_time() -> u64 {
...
+    storage: &Option<Arc<Storage>>,
+    namespace: &str,
+    key: &str,
+    parse_utf8: bool,
+    dg_name: &str,
+) -> Option<T> {
+    let storage = match storage {
+        None => return None,
+        Some(s) => s,
+    };
```
obelisk, line 271: let storage = storage?;

runtime/plaid/src/data/mod.rs
```diff
@@ -19,11 +19,16 @@ use std::{
...
 
+use serde::de::DeserializeOwned;
 use serde::Deserialize;
 use time::OffsetDateTime;
 
 pub use self::internal::DelayedMessage;
 
+const DATA_GENERATOR_STORAGE_PREFIX: &str = "__DATA_GENERATOR";
+const LAST_SEEN: &str = "last_seen";
+const ALREADY_SEEN_UUIDS: &str = "already_seen_uuids";
```
obelisk, line 30: These should end in `_KEY`

obelisk (changes requested):

runtime/plaid/src/data/mod.rs
michelemin: Fixed in 53974a6

PR #173, opened 2025-07-10
Implement S3 APIs

wbssbw:
This PR is a recreation of https://github.com/obelisk/plaid/pull/137

### DIFF
--- a/runtime/plaid-stl/src/aws/s3.rs
+++ b/runtime/plaid-stl/src/aws/s3.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use serde::{Deserialize, Serialize};
 
 use crate::PlaidFunctionError;
@@ -43,6 +45,17 @@ pub struct PutObjectRequest {
     pub object_key: String,
 }
 
+/// Request payload for tagging an object in S3.
+#[derive(Deserialize, Serialize)]
+pub struct PutObjectTagRequest {
+    /// The bucket name containing the object.
+    pub bucket_id: String,
+    /// Name of the object key.
+    pub object_key: String,
+    /// Tags to apply to the object
+    pub tags: HashMap<String, String>,
+}
+
 /// Represents an S3 object's metadata
 #[derive(Deserialize, Serialize)]
 pub struct ObjectAttributes {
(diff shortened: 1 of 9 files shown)


obelisk (dismissed):

runtime/plaid-stl/src/aws/s3.rs
```diff
@@ -0,0 +1,457 @@
...
+
+/// Request payload for retrieving an object from S3.
+#[derive(Deserialize, Serialize)]
+pub struct ListObjectsRequest {
+    /// The bucket name from which the object is requested.
+    pub bucket_id: String,
+    /// The key identifying the object to fetch.
+    pub prefix: String,
+    pub continuation_key: Option<String>,
+}
```
michelemin: Doc comments are wrong here: copy/paste? 😄

runtime/plaid-stl/src/aws/s3.rs
```diff
@@ -0,0 +1,457 @@
...
+/// - If bucket versioning is enabled, the operation inserts a delete marker, which becomes the current version of the object.
+///   To permanently delete an object in a versioned bucket, you must include the object’s `versionId` in the request.
+///   For more information about versioning-enabled buckets, see [Deleting object versions from a versioning-enabled bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeletingObjectVersions.html)
+/// - If bucket versioning is suspended, the operation removes the object that has a null `versionId`, if there is one, and
+///   inserts a delete marker that becomes the current version of the object. If there isn't an object with a null `versionId`,
+///   and all versions of the object have a `versionId`, Amazon S3 does not remove the object and only inserts a delete marker.
+///   To permanently delete an object that has a `versionId`, you must include the object’s `versionId` in the request.
+///   For more information about versioning-suspended buckets, see [Deleting objects from versioning-suspended buckets](https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeletingObjectsfromVersioningSuspendedBuckets.html)
```
michelemin: Easy...

runtime/plaid-stl/src/aws/s3.rs
```diff
@@ -0,0 +1,457 @@
...
+///   - A tag key can be up to 128 Unicode characters in length, and tag values can be up to 256 Unicode characters in length.
+///   - Neither keys nor values may begin with `aws:`. The set of allowed characters are
+///     letters (`a-z`, `A-Z`), numbers (`0-9`), and spaces representable in UTF-8, and the following characters: `+ - = . _ : / @`
+///   - At most 10 entries are allowed.
+///
+/// For full details on the S3 tagging model—including character limits, reserved prefixes,
+/// replication behavior, and how tags interact with bucket versioning—see the AWS S3
+/// Object Tagging documentation:
+/// <https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-tagging.html>
+pub fn put_object_tag(
```
michelemin: Nit
```suggestion
pub fn put_object_tags(
```

runtime/plaid-stl/src/aws/s3.rs
```diff
@@ -0,0 +1,457 @@
...
+/// Returns some or all (up to 1,000) of the objects in a bucket with each request.
+/// You can use the request parameters as selection criteria to return a subset of the objects in a bucket.
+/// See https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html for full documentation.
+///
+/// # Arguments
+///
+/// * `bucket_id` - The S3 bucket name or ARN where objects are stored.
+/// * `prefix` - Filters the listed objects to those with keys that begin with this prefix.
+/// * `continuation_key` - An optional token to continue pagination from a previous response.
+pub fn list_objects(
```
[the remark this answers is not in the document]
michelemin: I think we can keep it like this for now, given the limit of 1000 objects is anyway quite generous.

Then, if we think it's a problem, we can do something like the GH API, where we deprecate this function in favor of another one that handles pagination internally.

runtime/plaid/src/apis/aws/s3.rs
```diff
@@ -0,0 +1,485 @@
...
+    DeleteObjectError(SdkError<DeleteObjectError>),
+    ListObjectVersionsError(SdkError<ListObjectVersionsError>),
+    GetObjectAttributesError(SdkError<GetObjectAttributesError>),
+    BytesStreamError(aws_sdk_s3::primitives::ByteStreamError),
+    PresignError(aws_sdk_s3::presigning::PresigningConfigError),
+    NoContentLengthReturned,
+    ObjectTooLarge,
+}
+
+/// Configuration for a single S3 bucket, including permissions and the allowed rule.
```
michelemin: Nit
```suggestion
/// Configuration for a single S3 bucket, including permissions and the allowed rules.
```

runtime/plaid/src/apis/aws/s3.rs
```diff
@@ -0,0 +1,485 @@
...
+    pub async fn delete_object(
+        &self,
+        params: &str,
+        module: Arc<PlaidModule>,
+    ) -> Result<u32, ApiError> {
+        let request = serde_json::from_str::<DeleteObjectRequest>(params)
+            .map_err(|_| ApiError::BadRequest)?;
+
+        // Check that caller is allowed to write to this bucket
+        if !self.can_rule_access_bucket(module.clone(), &request.bucket_id, true) {
```
michelemin: I honestly don't know if cloning the entire `module` has a performance penalty. But we only care about its name, so we might as well pass something like `&module.name`.

runtime/plaid/src/apis/aws/s3.rs
```diff
@@ -0,0 +1,485 @@
...
+
+        if let Some(version) = request.version_id {
+            delete_request = delete_request.version_id(version)
+        }
+
+        delete_request
+            .send()
+            .await
+            .map_err(S3Errors::DeleteObjectError)?;
+
```
michelemin: I guess we are not logging a success because after the "attempting..." message we either see nothing (success) or we see an error in case of failure. But let's consider it...

runtime/plaid-stl/src/aws/s3.rs
```diff
@@ -0,0 +1,457 @@
...
+    let (expires_in, return_buffer_size) = match fetch_mode {
+        ObjectFetchMode::Presigned(duration) => (Some(duration), RETURN_BUFFER_SIZE),
+        _ => {
+            // We need to figure out how big our return buffer should be based on the response from get_object_attributes
+            let attributes = get_object_attributes(bucket_id, object_key)?;
+            let Some(ret_buffer_size) = attributes.object_size else {
+                return Err(PlaidFunctionError::InternalApiError);
+            };
+
+            (None, (ret_buffer_size * 5) as usize) // Bump the return buffer to account for how it will be serialized
```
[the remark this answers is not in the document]
michelemin: OK coming back after reading the runtime code. So the problem is that we JSON-serialize the bytes, right? In the sense that a single byte might become something like `123` (which is now 3 bytes). Yeah I see...

runtime/plaid/src/apis/aws/s3.rs
```diff
@@ -0,0 +1,485 @@
...
+        let tags = request
+            .tags
+            .into_iter()
+            .filter_map(|(key, value)| {
+                let key_len = key.len();
+                let val_len = value.len();
+
+                // Length checks (key: 1–128; value: 0–256)
+                if key_len < 1 || key_len > 128 || val_len > 256 {
+                    return None;
```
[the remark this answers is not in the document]
wbssbw: Yeah good call, will update the method to return the set of tags applied on success

PR #174, opened 2025-07-11
Handle empty values in storage providers

wbssbw:
This PR addresses an issue that can occur in the runtime when a storage key is present without a corresponding value. This was discovered in DynamoDB where it's possible to have items with a primary key but no other attributes. This vulnerability could have been exploited to disrupt a Plaid instance by maliciously inserting a key with no value to a storage provider, causing a denial of service.

To fix this, the StorageProvider trait has been updated to reflect that values can be optional. Specifically, `fetch_all` now returns a `Vec<(String, Option<Vec<u8>>)>`. This change has been propagated through the different `StorageProvider` implementations (`dynamodb`, `in_memory`, `sled`) and the call sites that use `fetch_all`.

Additionally, to reduce code duplication, default implementations for `get_namespace_byte_size` and `apply_migration` have been added to the `StorageProvider` trait. This was possible because all implementors of the trait had the same code for these functions. The new default implementations have been updated to handle the optional value from `fetch_all`.

### DIFF
--- a/runtime/plaid/src/bin/ddb_logback_explorer.rs
+++ b/runtime/plaid/src/bin/ddb_logback_explorer.rs
@@ -38,13 +38,12 @@ impl Ord for TmpLogback {
     }
 }
 
-impl TryFrom<(String, Vec<u8>)> for DbEntry {
+impl TryFrom<(String, Option<Vec<u8>>)> for DbEntry {
     type Error = String;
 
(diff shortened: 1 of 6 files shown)


obelisk (approved):

obelisk:
how is this handled at the API level? I would expect this to be a breaking change

wbssbw:
> how is this handled at the API level? I would expect this to be a breaking change

`fetch_all` isn't exposed to rules as a host function so no breaking changes

PR #178, opened 2025-07-24
Rewrite the cache sub-system to be more similar to storage

michelemin:


runtime/plaid/Cargo.toml
```diff
@@ -40,6 +40,7 @@ octocrab = "0.37"
 paste = "1.0"
 plaid_stl = { path = "../plaid-stl" }
 rcgen = { version = "0.10", features = ["x509-parser"] }
+redis = { version = "0.32.4", features = ["aio", "tokio-comp"], optional = true }
```
michelemin, line 43: TODO for me: Add feature test that uses `redis`.

runtime/plaid/src/cache/redis/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+        key: &str,
+        value: &str,
+    ) -> Result<Option<String>, CacheError> {
+        let mut connection = self
+            .client
+            .get_multiplexed_async_connection()
+            .await
+            .map_err(|e| {
+                CacheError::CacheAccessError(format!("Could not create redis connection: {e}"))
+            })?;
```
michelemin, line 93: To be checked if this is the optimal way, or if we can somehow persist the connection without having to take `&mut self`.

runtime/plaid/resources/config/cache.toml.redis
```diff
@@ -0,0 +1,11 @@
...
+
+[cache.cache_entries]
+default = 200
+[cache.cache_entries.log_type]
+[cache.cache_entries.module_overrides]
+
+[cache.backend]
+type = "Redis"
+hostname = "127.0.0.1"
+port = "6379"
```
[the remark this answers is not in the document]
michelemin: Added, together with tests against an instance that requires authentication

runtime/plaid/src/cache/redis/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+                    .arg(key)
+                    .arg(value)
+                    .arg(max_entries)
+                    .invoke_async(&mut connection)
+                    .await
+                    .map_err(|e| CacheError::PutError(e.to_string()))?;
+
+                if evicted {
+                    debug!("An item has been evicted from the Redis cache");
+                }
```
obelisk: I can see this being useful but I'm wondering if this is something that should be interrogated on the Redis side? Especially if collecting the information on if a key was evicted costs extra compute time

PR #183, opened 2025-08-07
Add APIs to add/remove repos to/from GH teams

michelemin:
Add APIs to add/remove repos to/from GH teams.

### DIFF
--- a/runtime/plaid/src/apis/github/teams.rs
+++ b/runtime/plaid/src/apis/github/teams.rs
@@ -8,6 +8,57 @@ use crate::{
     loader::PlaidModule,
 };
 
+use std::str::FromStr;
+
+enum Permission {
+    Admin,
+    Push,
+    Maintain,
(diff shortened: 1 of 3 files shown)


runtime/plaid/src/apis/github/teams.rs
```diff
@@ -8,6 +8,57 @@ use crate::{
...
+        let s = match self {
+            Permission::Admin => "admin",
+            Permission::Push => "push",
+            Permission::Maintain => "maintain",
+            Permission::Triage => "triage",
+            Permission::Pull => "pull",
+        };
+        write!(f, "{}", s)
+    }
+}
```
obelisk, line 47: This won't support custom roles. We may want that

obelisk: But we don't have to do it now.

obelisk (approved):

PR #185, opened 2025-08-08
Add blocksmith tool and AES runtime API

michelemin:
This PR does the following:
* Add a simple tool for encrypting / decrypting strings under AES-128-CBC.
* Add an AES runtime API for rules to encrypt / decrypt with keys defined in Plaid's config. The config supports specifying which rules can perform which actions with a given key.

### DIFF
--- a/runtime/plaid/Cargo.toml
+++ b/runtime/plaid/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "plaid"
-version = "0.26.0"
+version = "0.27.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/modules/Cargo.lock
+++ b/modules/Cargo.lock
@@ -199,7 +199,7 @@ checksum = "57c0d7b74b563b49d38dae00a0c37d4d6de9b432382b2892f0574ddcae73fd0a"
 
 [[package]]
 name = "plaid_stl"
-version = "0.26.0"
+version = "0.27.0"
 dependencies = [
  "base64 0.13.1",
  "chrono",
--- a/runtime/Cargo.lock
+++ b/runtime/Cargo.lock
@@ -3224,7 +3224,7 @@ checksum = "7edddbd0b52d732b21ad9a5fab5c704c14cd949e5e9a1ec5929a24fded1b904c"
 
 [[package]]
 name = "plaid"
-version = "0.26.0"
+version = "0.27.0"
 dependencies = [
  "aes",
  "alkali",
@@ -3278,7 +3278,7 @@ dependencies = [
 
 [[package]]
 name = "plaid_stl"
-version = "0.26.0"
+version = "0.27.0"
 dependencies = [
  "base64 0.13.1",
  "chrono",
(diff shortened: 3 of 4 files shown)


runtime/plaid/Cargo.toml
```diff
@@ -19,6 +19,7 @@ aws-sdk-dynamodb = { version = "1.69.0", optional = true }
 aws-sdk-kms = { version = "1.41.0", optional = true }
 aws-sdk-secretsmanager = "1.57.0"
 base64 = "0.13"
+blocksmith = { path = "../blocksmith" }
```
obelisk, line 22: Your code is in the wrong place. The code should be in the runtime and then blocksmith should import from that (at least that's what makes the most sense to me unless we want to start breaking functionality out into subcrates).

runtime/plaid/src/apis/aes/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+    ) -> bool {
+        self.key_specs
+            .get(&key_id.to_string())
+            .and_then(|key_spec| key_spec.rules_and_actions.get(&module.to_string()))
+            .and_then(|allowed_actions| Some(allowed_actions.contains(&action)))
+            == Some(true)
+    }
+
+    /// Perform an AES encryption using a key defined in Plaid's config.
+    pub async fn encrypt_local(
```
obelisk, line 62: Why is this called `encrypt_local` and further why is it async?

runtime/plaid/src/apis/aes/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+                .map_err(|_| ApiError::AesError("Failed to encrypt plaintext".to_string()))
+        } else {
+            Err(ApiError::AesError(
+                "Missing key or operation not permitted".to_string(),
+            ))
+        }
+    }
+
+    /// Perform an AES decryption using a key defined in Plaid's config.
+    pub async fn decrypt_local(
```
obelisk, line 92: Same question

runtime/plaid/src/apis/mod.rs
```diff
@@ -1,3 +1,4 @@
+pub mod aes;
```
obelisk, line 1: I would probably create a cryptography module and move this into that as a submodule.

runtime/plaid/Cargo.toml
michelemin: I also don't mind the idea of subcrates, which are smaller and easier to test, if the functionality they offer is well scoped.

runtime/plaid/src/apis/mod.rs
michelemin: Done!

runtime/plaid/src/apis/cryptography/aes.rs
```diff
@@ -0,0 +1,154 @@
...
+        if self.aes.is_none() {
+            return Err(ApiError::CryptographyError(
+                "API not configured".to_string(),
+            ));
+        }
+        // We have an AES API, so we can continue
+
+        let payload: AesDecryptPayload = serde_json::from_str(&params)
+            .map_err(|_| ApiError::CryptographyError("Failed to parse payload".to_string()))?;
+        if self.can_module_perform_aes_action(&module.name, &payload.key_id, AesAction::Decrypt) {
```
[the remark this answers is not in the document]
michelemin: Done!

obelisk (approved):

PR #189, opened 2025-08-15
SQS Data Generator 

harry-anderson:
data generator for reading messages from AWS SQS queue

### DIFF
--- a/runtime/plaid/src/data/sqs/mod.rs
+++ b/runtime/plaid/src/data/sqs/mod.rs
@@ -26,6 +26,10 @@ pub struct SQSConfig {
     #[serde(default = "default_sleep")]
     #[serde(deserialize_with = "parse_duration")]
     pub sleep_duration: Duration,
+    #[serde(default = "default_max_num_messages")]
+    pub max_num_messages: i32,
+    #[serde(default = "default_wait_time_seconds")]
+    pub wait_time_seconds: i32,
 }
 
 /// This function provides the default sleep duration.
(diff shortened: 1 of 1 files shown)


michelemin (approved):

runtime/plaid/src/data/sqs/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub queue_url: String,
+    /// How to authenticate to AWS
+    pub authentication: AwsAuthentication,
+    /// Denotes if logs produced by this generator are allowed to initiate log backs
+    #[serde(default)]
+    logbacks_allowed: LogbacksAllowed,
+    /// Number of milliseconds to wait in between calls to the API.
+    /// If no value is provided here, we will use a default value (10 second).
+    #[serde(default = "default_sleep_milliseconds")]
+    pub sleep_duration: u64,
```
wbssbw: Let's use the `parse_duration` deserializer from the websocket data generator here (will probably need to pull it into utils or something)

wbssbw: And then have `default_sleep_milliseconds` return a `Duration` instead of `u64`

runtime/plaid/src/data/sqs/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+        trace!("sqs/{} draining queue", self.config.name);
+
+        loop {
+            // poll the SQS queue
+            let res = self
+                .client
+                .receive_message()
+                .queue_url(&self.config.queue_url)
+                .max_number_of_messages(10) // just get max if available
+                .wait_time_seconds(1) // no long polling
```
wbssbw: We may want to make these configurable - I don't see any reason why we'd want to pull less than the max, but doesn't hurt to give the option

michelemin: Done here: [Make SQS more configurable](https://github.com/obelisk/plaid/pull/189/commits/d17266092c145730c7e7e8717d22dc822fb1ba89)

michelemin (approved):

PR #191, opened 2025-08-25
Add runtime API to check a repo's CODEOWNERS file

michelemin:
The API for modules will return whether a CODEOWNERS file is:
* OK - It is present and has no errors
* Missing
* Invalid - It is present but there is at least one error

### DIFF
--- a/runtime/plaid-stl/Cargo.toml
+++ b/runtime/plaid-stl/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "plaid_stl"
-version = "0.28.0"
+version = "0.27.2"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/runtime/plaid-stl/Cargo.toml
+++ b/runtime/plaid-stl/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "plaid_stl"
-version = "0.27.1"
+version = "0.28.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/runtime/plaid-stl/src/github/mod.rs
+++ b/runtime/plaid-stl/src/github/mod.rs
@@ -158,33 +158,33 @@ pub struct CheckCodeownersParams {
     pub repo: String,
 }
 
+/// An error detected in a CODEOWNERS file.
+/// See https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-codeowners-errors for more details.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct CodeownersError {
+    pub kind: String,
+    pub message: String,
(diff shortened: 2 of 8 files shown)


runtime/plaid-stl/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "plaid_stl"
-version = "0.27.1"
+version = "0.28.0"
```
obelisk, line 3: Technically this is backwards compatible so could be 27.2.

I don't feel strongly though.

runtime/plaid-stl/src/github/mod.rs
```diff
@@ -151,6 +151,51 @@ pub struct RepositoryCustomProperty {
...
 
+/// Parameters sent to the runtime when checking a repo's CODEOWNERS file.
+#[derive(Serialize, Deserialize)]
+pub struct CheckCodeownersParams {
+    pub owner: String,
+    pub repo: String,
+}
+
+/// Status for a repo's CODEOWNERS file
+pub enum CodeownersStatus {
```
obelisk, line 162: This should return the vector of errors in the `Invalid` case. This will surely be needed at some point and prevents us from needing to make a backwards incompatible change.

If you feel that you don't want to expose that much information, you could have two APIs :)

runtime/plaid-stl/src/github/mod.rs
```diff
@@ -151,6 +151,51 @@ pub struct RepositoryCustomProperty {
...
+    Invalid,
+}
+
+impl From<String> for CodeownersStatus {
+    fn from(value: String) -> Self {
+        match value.as_str() {
+            "Ok" => Self::Ok,
+            "Missing" => Self::Missing,
+            "Invalid" => Self::Invalid,
+            _ => unreachable!(), // we are receiving this string from the runtime, so it should never happen
```
obelisk, line 177: This is true but you could use this on strings that didn't come from the runtime accidentally. Thus this should probably be `TryFrom`.

runtime/plaid-stl/src/github/mod.rs
```diff
@@ -1902,3 +1947,47 @@ pub fn remove_repo_from_team(
...
+    }
+
+    return_buffer.truncate(res as usize);
+
+    // This should be safe because unless the Plaid runtime is expressly trying
+    // to mess with us, this came from a String in the API module.
+    let response_body =
+        String::from_utf8(return_buffer).map_err(|_| PlaidFunctionError::InternalApiError)?;
+
+    Ok(CodeownersStatus::from(response_body))
```
obelisk, line 1992: Yeah this should definitely be fallible that way mixing versions of Plaid rules and runtimes is more easily debuggable.

obelisk (changes requested):

runtime/plaid-stl/src/github/mod.rs
michelemin: Fair point 😄

PR #202, opened 2025-10-07
Make JWT subsystem safer and add tests

michelemin:
* Rely on a more prescriptive data type to pass JWT requests, which enforces presence of some key fields
* Support extra fields and headers, but only if allow-listed in the config
* Add test rule

Open points:
- [x] `exp`: configure max TTL at key level
- [x] `iat`: flag to enforce accurate `iat`. If that's turned on, set `iat` to now
- [x] Have `enforced_aud: Option<String>` at key level
- [x] Support allow-listed headers at key level
- [x] Extend tests

### DIFF
--- a/modules/tests/test_jwt/harness/harness.sh
+++ b/modules/tests/test_jwt/harness/harness.sh
@@ -20,7 +20,7 @@ sleep 2
 
 kill $RH_PID 2>&1 > /dev/null
 
-echo -e "OK\nOK\nOK\nOK\nOK\nOK\nOK" > expected.txt
+echo -e "OK\nOK\nOK\nOK\nOK\nOK\nOK\nOK" > expected.txt
 diff expected.txt $FILE
 RESULT=$?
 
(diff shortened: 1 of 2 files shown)


runtime/plaid/src/apis/web/mod.rs
```diff
@@ -93,31 +87,47 @@ impl Web {
             )));
         }
 
+        // TODO Discuss with @obelisk if we want to validate fields like 'iat' and 'exp'
+
+        let mut claims = HashMap::<String, Value>::new();
+        claims.insert("sub".to_string(), Value::String(request.sub.clone()));
+        claims.insert("iat".to_string(), Value::Number(request.iat.into()));
+        claims.insert("exp".to_string(), Value::Number(request.exp.into()));
```
obelisk, line 95: I would configure at the key level what the maximum TTL is

runtime/plaid/src/apis/web/mod.rs
```diff
@@ -93,31 +87,47 @@ impl Web {
...
+
+        // Include extra fields, but only if they are allowlisted in the config
+        if let Some(extra_fields) = request.extra_fields {
+            if let Some(allowlisted_extras) = &self.config.allowlisted_extra_fields {
```
obelisk, line 99: Would also add at the key level extra fields that are allowed in the header and body (independently).

obelisk (changes requested):

PR #210, opened 2025-10-21
Add GitHub API function to add labels to pull requests

wbssbw:
## Summary

This PR introduces a new GitHub API function that enables Plaid modules to add labels to pull requests and issues. 

## Changes

- **STL**: Added `AddLabelsRequest` struct and `add_labels` function in `runtime/plaid-stl/src/github/pull_request.rs`
- **Runtime**: Implemented `add_labels` method in GitHub API handler in `runtime/plaid/src/apis/github/pull_requests.rs`
- **Function Registration**: Added function mapping and implementation in `runtime/plaid/src/functions/api.rs`
- **Documentation**: Fixed return type documentation for `create_pull_request` function

## Rationale

This functionality was needed to enable automated labeling of pull requests and issues from Plaid modules. The implementation uses the GitHub Issues API endpoint `/repos/{owner}/{repo}/issues/{number}/labels` which works for both pull requests and issues, following GitHub's API design where pull requests are treated as a special type of issue.

### DIFF
--- a/modules/Cargo.lock
+++ b/modules/Cargo.lock
@@ -199,7 +199,7 @@ checksum = "57c0d7b74b563b49d38dae00a0c37d4d6de9b432382b2892f0574ddcae73fd0a"
 
 [[package]]
 name = "plaid_stl"
-version = "0.29.0"
+version = "0.29.1"
 dependencies = [
  "base64 0.13.1",
  "chrono",
--- a/runtime/Cargo.lock
+++ b/runtime/Cargo.lock
@@ -3307,7 +3307,7 @@ checksum = "7edddbd0b52d732b21ad9a5fab5c704c14cd949e5e9a1ec5929a24fded1b904c"
 
 [[package]]
 name = "plaid"
-version = "0.29.0"
+version = "0.29.1"
 dependencies = [
  "aes",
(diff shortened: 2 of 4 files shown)


runtime/plaid/src/apis/github/pull_requests.rs
```diff
@@ -154,6 +154,41 @@ impl Github {
...
+
+        let owner = self.validate_org(&request.owner)?;
+        let repo = self.validate_repository_name(&request.repo)?;
+
+        info!(
+            "Adding labels to issue/PR #{} in [{owner}/{repo}] org on behalf of {module}",
+            request.number
+        );
+
+        let address = format!("/repos/{owner}/{repo}/issues/{}/labels", request.number);
```
michelemin: The `number` and `labels` are not validated, and I wonder if this has consequences.

For `number` I am not so worried because it's a number (and it must be, otherwise the deserialization would fail).

I am more worried about `labels` which is a vector of strings. But I don't see an immediate way for the caller to abuse this: the URL does not depend on those, so if the body is wrong (e.g., contains strange characters) we rely on the GH API to reject this. So I think it's fine... 🤔

PR #214, opened 2025-11-07
Enforce max message and frame size at protocol level

wbssbw:
## Summary

Configures WebSocket connections to enforce message and frame size limits at the protocol level instead of after reading messages into memory. This prevents potential memory exhaustion attacks and improves resource management by rejecting oversized messages before they consume application memory.

## Changes

- **WebSocket Configuration**: Added `max_frame_size` configuration option alongside existing `max_message_size`
- **Protocol-Level Enforcement**: Modified `establish_connection` to use `connect_async_with_config` with `WebSocketConfig` to enforce limits at the protocol level
- **Removed Redundant Checks**: Eliminated post-read message size validation since the protocol now handles rejection automatically
- **Crypto Provider**: Added rustls crypto provider initialization to support the new WebSocket configuration API
- **Dependencies**: Added explicit rustls dependency with ring crypto provider support

## Rationale

The previous implementation read entire WebSocket messages into memory before checking size limits, which defeated the purpose of having size restrictions for memory protection. By configuring limits at the protocol level using `WebSocketConfig`, oversized messages are rejected during the WebSocket handshake or frame processing, preventing memory consumption.

## Testing

The changes maintain backward compatibility with existing configurations. WebSocket connections with oversized messages will now fail at the protocol level instead of being processed and then discarded.

### DIFF
--- a/runtime/plaid/src/data/websocket/mod.rs
+++ b/runtime/plaid/src/data/websocket/mod.rs
@@ -253,8 +253,10 @@ struct WebSocketClient {
     /// Manages a list of URI entries and handles the selection and retry logic for connection attempts.
     uri_selector: UriSelector,
     /// The maximum size of a message received from a socket that will be processed.
-    /// If not specified, the default value from `default_message_size` is used (`64 kB`)
+    /// If not specified, the default value from `default_message_size` is used (`64 KiB`)
     max_message_size: usize,
+    /// The maximum size of a frame received from a socket that will be processed.
+    /// If not specified, the default value from `default_frame_size` is used (`16 KiB`)
     max_frame_size: usize,
 }
 


runtime/plaid/src/data/websocket/mod.rs
```diff
@@ -236,6 +255,7 @@ struct WebSocketClient {
     /// The maximum size of a message received from a socket that will be processed.
     /// If not specified, the default value from `default_message_size` is used (`64 kB`)
     max_message_size: usize,
+    max_frame_size: usize,
```
michelemin: Nit - missing doc string.

runtime/plaid/Cargo.toml
```diff
@@ -61,6 +66,7 @@ reqwest = { version = "0.12", default-features = false, features = [
     "cookies",
 ] }
 reqwest_cookie_store = "0.8.0"
+rustls = { version = "0.23.17", features = ["ring"] }
```
michelemin: Let's see what @obelisk says about this.
In case, would we have a "manual" alternative, like "stream-read from the frame" (if that even makes sense) and count the bytes, rather than relying on `connect_async_with_config` which iiuc is forcing us to add this?

wbssbw: Doesn't seem like there's any alternative unfortunately

PR #215, opened 2025-11-09
Google Docs API for rules

harry-anderson:
Google docs + Google Drive API for rules

### Authentication
Plaid uses Google OAuth Token Grant Authentication to interact with Google APIs. What this means is that plaid uses the same type of authentication as a SaaS application like Zapier would use. 

1. Google OAuth Client - The Plaid admin creates a Google Cloud Project with Google Docs + Google Drive APIs enabled. Google OAuth Client and saves `client_id` and `client_secret` for use later in (2)
2. Google account (puppet) OAuth refresh token - Plaid requires a google account to impersonate. This can be any google account: personal or google workspace. So the plaid admin, using a puppet google account goes through OAuth Grant Flow (using OAuth Credentials from (1)). The scopes for the OAuth Grant are Read/Write access to google docs / google drive for the puppet google account. The result of the OAuth Grant is an `access_token` (short lived) and a `refresh_token` (long lived) . The refresh token is stored for use later to obtain a fresh access token
3. Plaid admin configures plaid with the following secrets: `client_id`, `client_secret`, `refresh_token` (for the puppet account). Then plaid can interact with google APIs on-behalf of the puppet account - for example upload documents to google drive folders with the same level of access as the puppet account

### Multipart Upload

This API does not directly use the Google Docs API. Instead it uses the Google Drive API which has support for document conversion. This gives much more flexibility in the manipulation of files and contents. So when sending a multipart upload request we tell google what is the current mime type and what is the mime type that we would like to upload, Google does the conversion for us.

### Granular Permissions
Google's OAuth Grant Authentication Scope is broad - everything the puppet account has access to. 
We want to impose a secondary layer of permissions so rules cannot read/write any abitary google doc / drive folder that the puppet account has access to. So we maintain a mapping for 'readers' and 'writers' similar to other APIs. Where for a given google drive folder_id a list of rules is allowed to read/write.

### Tests
```bash
cd plaid/runtime/plaid
cargo test apis::gcp::google_docs::tests::create_markdown_doc -- --nocapture  
cargo test apis::gcp::google_docs::tests::create_csv_sheet -- --nocapture
cargo test apis::gcp::google_docs::tests::permission_checks -- --nocapture

### OAuth Flow for Refresh Token
cargo test apis::gcp::google_docs::tests::get_refresh_token -- --nocapture
```

### DIFF
--- a/runtime/plaid/src/apis/gcp/google_docs.rs
+++ b/runtime/plaid/src/apis/gcp/google_docs.rs
@@ -278,7 +278,7 @@ impl GoogleDocs {
 /// utility function for converting markdown into HTML
 fn markdown_to_html(md: &str) -> String {
     // Convert Markdown to HTML
-    // We enable folders and footnotes for better compatibility
+    // We enable tables and footnotes for better compatibility
     let mut options = Options::empty();
     options.insert(Options::ENABLE_TABLES);
     options.insert(Options::ENABLE_FOOTNOTES);
--- a/.github/workflows/FeatureCheck.yml
+++ b/.github/workflows/FeatureCheck.yml
@@ -2,9 +2,9 @@ name: Feature Checks
 
 on:
   push:
-    branches: [ main, develop ]
+    branches: [main, develop]
   pull_request:
-    branches:  "**"
+    branches: "**"
 
 permissions:
   contents: read
@@ -16,7 +16,15 @@ jobs:
   build-feature-combination:
     strategy:
       matrix:
-        features: ["sled,cranelift", "aws,cranelift", "aws,sled,cranelift", "aws,cranelift,redis"]
+        features:
+          [
+            "sled,cranelift",
+            "aws,cranelift",
+            "aws,sled,cranelift",
+            "aws,cranelift,redis",
+            "aws,gcp,sled,cranelift",
+            "aws,gcp,cranelift,redis",
+          ]
     runs-on: ubuntu-latest
     steps:
       - name: Setup Rust
--- a/runtime/plaid/src/functions/api.rs
+++ b/runtime/plaid/src/functions/api.rs
@@ -498,14 +498,14 @@ impl_new_sub_module_function_with_error_buffer!(
     gcp,
     google_docs,
     create_doc_from_markdown,
-    ALLOW_IN_TEST_MODE
+    DISALLOW_IN_TEST_MODE
 );
 #[cfg(feature = "gcp")]
 impl_new_sub_module_function_with_error_buffer!(
     gcp,
     google_docs,
     create_sheet_from_csv,
-    ALLOW_IN_TEST_MODE
+    DISALLOW_IN_TEST_MODE
 );
 
 // Npm Functions


runtime/plaid/src/apis/gcp/google_docs.rs
```diff
@@ -0,0 +1,641 @@
...
+    "https://www.googleapis.com/upload/drive/v3/files?uploadType=multipart";
+
+impl GoogleDocs {
+    pub fn new(config: GoogleDocsConfig) -> Self {
+        let GoogleDocsConfig {
+            client_id,
+            client_secret,
+            refresh_token,
+            r,
+            rw: wr,
```
harry-anderson: fix

runtime/plaid/src/apis/gcp/google_docs.rs
```diff
@@ -0,0 +1,641 @@
...
+    options.insert(Options::ENABLE_FOOTNOTES);
+
+    let parser = Parser::new_ext(md, options);
+    let mut output = String::new();
+    html::push_html(&mut output, parser);
+
+    output
+}
+
+/// untility function for rendering static template
```
harry-anderson: typo

runtime/plaid/src/apis/gcp/google_docs.rs
```diff
@@ -0,0 +1,639 @@
...
+            .and_then(|v| v.as_u64())
+            .unwrap_or(3600);
+
+        let expiry = Instant::now() + Duration::from_secs(expires_in.saturating_sub(60));
+        *lock = Some((access_token.clone(), expiry));
+
+        Ok(access_token)
+    }
+
+    /// Uploads a file to a Google Drive folder using multipart uploads
```
harry-anderson: upload

runtime/plaid/src/apis/aws/dynamodb.rs
```diff
@@ -38,7 +38,7 @@ pub struct DynamoDbConfig {
 /// Represents the DynamoDB API client.
 /// NOTE: if Plaid is configured with the DynamoDB database backend, sharing tables here will lead to undefined behaviour
 pub struct DynamoDb {
-    /// The underlying client used to interact with the KMS API.
+    /// The underlying client used to interact with the Dynamodb API.
```
obelisk: 👍

runtime/plaid/src/apis/gcp/google_docs.rs
```diff
@@ -0,0 +1,639 @@
...
+
+impl GoogleDocs {
+    pub fn new(config: GoogleDocsConfig) -> Self {
+        let GoogleDocsConfig {
+            client_id,
+            client_secret,
+            refresh_token,
+            r,
+            rw,
+        } = config;
```
obelisk: You could also just store the config inside the `GoogleDocs` struct which will then keep them in sync if you ever need to add anything

runtime/plaid/src/apis/gcp/google_docs.rs
```diff
@@ -0,0 +1,639 @@
...
+            )));
+        }
+
+        let json: Value = response.json().await?;
+
+        json.get("id")
+            .and_then(|v| v.as_str())
+            .map(|s| s.to_string())
+            .ok_or(GoogleDocsError::MissingField("id"))
+    }
```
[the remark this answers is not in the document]
harry-anderson: These functions require access to private members on the struct

runtime/plaid/src/apis/gcp/google_docs.rs
```diff
@@ -0,0 +1,549 @@
...
+
+        let output = CreateDocFromMarkdownOutput { document_id };
+        serde_json::to_string(&output).map_err(|err| ApiError::SerdeError(err.to_string()))
+    }
+}
+
+/// utility function for converting markdown into HTML
+fn markdown_to_html(md: &str) -> String {
+    // Convert Markdown to HTML
+    // We enable folders and footnotes for better compatibility
```
michelemin: Nit
```suggestion
    // We enable tables and footnotes for better compatibility
```
?

michelemin (dismissed):
LGTM, maybe add something to `FeatureCheck.yml` but I would not block the PR on this.

runtime/plaid/src/apis/gcp/google_docs.rs
```diff
@@ -0,0 +1,549 @@
...
+            .and_then(|v| v.as_str())
+            .map(|s| s.to_string())
+            .ok_or(GoogleDocsError::MissingField("access_token"))?;
+
+        let expires_in = json
+            .get("expires_in")
+            .and_then(|v| v.as_u64())
+            .unwrap_or(3600);
+
+        let expiry = Instant::now() + Duration::from_secs(expires_in.saturating_sub(60));
```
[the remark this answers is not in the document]
harry-anderson: yep

PR #218, opened 2025-11-10
Add non-blocking Splunk logging

wbssbw:
## Summary

Introduces a non-blocking variant of the Splunk HEC logging API that allows high-throughput logging without blocking the the response. The existing blocking behavior is preserved for cases where delivery confirmation is required, while the new non-blocking API queues requests and executes them asynchronously in the background.

## Changes

- **STL API**: Added `post_log_non_blocking()` function that returns immediately after queuing the request
- **Runtime implementation**: Added blocking/non-blocking logic with separate handler methods for each execution mode
- **Request structure**: Replaced HashMap-based parameters with structured `PostLogRequest` containing blocking flag
- **Documentation**: Added comprehensive doc comments explaining behavior, parameters, and use cases for both APIs
- **Logging**: Updated log messages to clearly indicate blocking vs non-blocking operation mode

## Rationale

High-throughput logging scenarios require non-blocking operations to avoid performance bottlenecks. The existing blocking API creates latency when Splunk is slow or unavailable. This change provides both options: blocking for critical logs requiring delivery confirmation, and non-blocking for high-volume logs where performance is prioritized over delivery guarantees.

## Performance/Operational Impact

The non-blocking API improves performance for high-throughput logging by eliminating network wait times. Background tasks handle HTTP requests asynchronously using `tokio::spawn`. Failed deliveries are logged internally but don't impact caller performance. No breaking changes to existing blocking API usage.

### DIFF
--- a/runtime/plaid-stl/src/splunk/mod.rs
+++ b/runtime/plaid-stl/src/splunk/mod.rs
@@ -1,9 +1,32 @@
-use std::collections::HashMap;
-
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 
 use crate::PlaidFunctionError;
 
+#[derive(Serialize, Deserialize)]
+pub struct PostLogRequest {
+    pub hec_name: String,
+    pub data: String,
+    pub blocking: bool,
+}
+
+/// Sends a log message to Splunk HEC using a blocking HTTP request.
+///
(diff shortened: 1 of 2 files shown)


obelisk:
How do you handle someone spamming this API and taking down the entire platform?

@copilot

PR #221, opened 2025-11-18
DynamoDB improvements

harry-anderson:
Improvements to the dynamodb api for rules:
- address comments from @obelisk https://github.com/obelisk/plaid/pull/190#pullrequestreview-3466703509
- revise docs for dynamodb and remove inline tags
- Add `reserved_tables` config to prevent rules from accessing 'reserved_tables' i.e dynamodb table used by the storage API

### DIFF
--- a/runtime/plaid/src/apis/aws/dynamodb.rs
+++ b/runtime/plaid/src/apis/aws/dynamodb.rs
@@ -31,7 +31,6 @@ pub struct DynamoDbConfig {
     /// Configured readers - maps a table name to a list of rules that are allowed to READ data
     r: HashMap<String, HashSet<String>>,
     /// Reserved tables - list of 'reserved' table names which rules cannot access
-    /// For the purpose of preventing rules rule accessing 'storage' tables in
     #[serde(default)]
     reserved_tables: Option<HashSet<String>>,
 }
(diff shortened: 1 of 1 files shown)


runtime/plaid/src/apis/aws/dynamodb.rs
```diff
@@ -83,15 +98,34 @@ impl DynamoDb {
...
+        Self {
+            client,
+            r,
+            rw,
+            reserved_tables,
+        }
     }
 
+    /// Checks if module can perform given give action
+    /// Modules are registerd as as read (R) or write (RW) under self.
```
michelemin: Typos
```suggestion
    /// Checks if module can perform a given action
    /// Modules are registered as as read (R) or write (RW) under self.
```

runtime/plaid/src/apis/aws/dynamodb.rs
```diff
@@ -117,6 +151,17 @@ impl DynamoDb {
...
+                if let Some(table_readers) = self.r.get(table_name) {
+                    // check if this module has read access to this table
+                    if table_readers.contains(&module.to_string()) {
+                        warn!(
+                            "[{module}] trying to [write] but only has [read] permission for dynamodb table [{table_name}]"
+                        );
+                        return Err(ApiError::BadRequest);
+                    }
+                }
+
```
michelemin: I would put this block _after_ checking for Write access.

Otherwise, imagine that a module appears both in the `r` and `rw` sections for a table: we would log a warning but then everything would be fine.

So I'd say we first check the happy path: if something goes wrong, we dig deeper and try to understand what happened.

runtime/plaid/src/apis/aws/dynamodb.rs
```diff
@@ -30,6 +30,10 @@ pub struct DynamoDbConfig {
     rw: HashMap<String, HashSet<String>>,
     /// Configured readers - maps a table name to a list of rules that are allowed to READ data
     r: HashMap<String, HashSet<String>>,
+    /// Reserved tables - list of 'reserved' table names which rules cannot access
+    /// For the purpose of preventing rules rule accessing 'storage' tables in
```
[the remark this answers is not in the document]
harry-anderson: done

PR #225, opened 2025-12-04
[ON HOLD] Parse Content-Length header if present, otherwise stop processing

michelemin:
Because some requests might reach warp without a `Content-Length` header and that would be breaking the filter we had in place, causing warp to respond with a HTTP 411. See [here](https://docs.rs/warp/latest/warp/filters/body/fn.content_length_limit.html) for more info.

Instead, now we parse the `Content-Length` header if present and manually check it's below a threshold, but we reply HTTP 200 in all cases.


obelisk:
Closing in favour of #227

PR #226, opened 2025-12-04
[GH API] Support custom media types when fetching files

michelemin:
Documentation: https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content

We now support `raw`, `html` and `object` on top of the default media type. Choosing a media type results in setting the appropriate `Accept` header when making the GET call to the REST API.

Changes are backward compatible for rules.

### DIFF
--- a/modules/Cargo.lock
+++ b/modules/Cargo.lock
@@ -199,7 +199,7 @@ checksum = "57c0d7b74b563b49d38dae00a0c37d4d6de9b432382b2892f0574ddcae73fd0a"
 
 [[package]]
 name = "plaid_stl"
-version = "0.33.1"
+version = "0.33.2"
 dependencies = [
  "base64 0.13.1",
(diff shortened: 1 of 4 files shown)


runtime/plaid/src/apis/github/mod.rs
```diff
@@ -120,6 +120,36 @@ impl Github {
         }
     }
 
+    /// Make a GET request with custom headers to the GitHub API using the GitHub app library.
+    /// Note - This function does not do any validation on the provided headers. That's because
+    /// it's not exposed to the rules but only callable from within the runtime itself. Therefore
+    /// we assume that all necessary validation has already been performed by the calling function.
+    async fn make_get_request_with_headers(
```
obelisk: Should the generic one call into this but just pass no headers?

PR #227, opened 2025-12-06
Fix Webhook POST For Better Content-Length Support

obelisk:
We've been experiencing some issues with some data providers failing health checks against Plaid because they use zero length POST requests to our webhooks. Since we enforced a content length using a `warp` filter, if these requests don't denote their zero length (which they often don't) we will reject the request.

This leads to the data providers thinking our service is down and disabling the sending of logs to us, even though if they actually sent logs, it would work.

This PR attempts to fix this by removing the `warp` filter and replacing it with custom streaming and byte counting. This allows us to handle there being no content-length provided while still not blindly allowing unlimited data to be buffered in the runtime.

I also added a new simplified system for running Plaid on a local machine to test things such as this.


michelemin (approved):
Very nice and loving the `jrp`!

PR #230, opened 2025-12-09
Fix WebSocket URI selector to prevent unsafe heap assumptions

wbssbw:
## Summary

Refactors the `UriSelector` to use a handle-based approach. Previously, the selector implicitly assumed the URI at the top of the heap remained unchanged between `next_uri()` and subsequent `mark_failed()` or `reset_failure()` calls, creating fragility under future modifications. The new design requires callers to pass back the exact `UriEntry` they received, eliminating this unsafe assumption while also enabling proper failure tracking for connections that drop after being established.

## Changes

**Code (`runtime/plaid/src/data/websocket/selector.rs`)**
- Changed `next_uri()` to pop and return `UriEntry` instead of `(String, Uri)` tuple
- Updated `mark_failed()` to accept `UriEntry` by value for reinsertion
- Updated `reset_failure()` to accept and return `UriEntry` without reinserting (caller holds it until failure)
- Removed implicit heap operations; entries now explicitly passed between selector and caller

**Code (`runtime/plaid/src/data/websocket/mod.rs`)**
- Updated `start()` to return `Option<UriEntry>` instead of `Option<String>`
- Refactored connection flow to pass `UriEntry` through success and failure paths

## Rationale

The original design relied on implicit ordering guarantees that could be violated by future code changes. The handle-based approach makes the contract explicit: the caller receives an entry, uses it, and returns it for state updates. This eliminates silent bugs where the wrong URI could be penalized if heap state changed between operations.

### DIFF
--- a/runtime/plaid/src/data/websocket/mod.rs
+++ b/runtime/plaid/src/data/websocket/mod.rs
@@ -276,6 +276,7 @@ impl WebSocketClient {
             configuration.uris.clone(),
             configuration.min_retry_duration,
             configuration.max_retry_duration,
+            configuration.log_type.clone(),
         );
 
         Self {
--- a/runtime/plaid/src/data/websocket/selector.rs
+++ b/runtime/plaid/src/data/websocket/selector.rs
@@ -1,4 +1,5 @@
 use http::Uri;
+use log::debug;
 use std::{
     collections::{BinaryHeap, HashMap},
     time::{Duration, Instant},
@@ -51,6 +52,8 @@ pub struct UriSelector {
     initial_retry_after: Duration,
     /// The maximum duration to wait before retrying a connection.
     max_retry_after: Duration,
+    /// The type of logs the websockets in this selector produce
+    log_type: String,
 }
 
 impl Ord for UriEntry {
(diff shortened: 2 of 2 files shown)


runtime/plaid/src/data/websocket/mod.rs
```diff
@@ -222,19 +222,22 @@ impl WebsocketGenerator {
...
+
                     if let Err(e) = logger.log_websocket_dropped(socket_name.clone()) {
                         error!(
                             "Failed to log WebSocket drop for socket [{socket_name}]. Error: {e}",
                         );
                     };
 
-                    client.uri_selector.mark_failed();
+                    // Mark the URI as failed since the connection dropped after being established
+                    client.uri_selector.mark_failed(uri_entry);
```
obelisk: Why the change?

runtime/plaid/src/data/websocket/mod.rs
```diff
@@ -302,12 +306,16 @@ impl WebSocketClient {
...
+    /// - Returns the UriEntry so the caller can mark it as failed if needed.
+    async fn start(&mut self) -> Option<UriEntry> {
+        let Some(uri_entry) = self.uri_selector.next_uri().await else {
             error!("No URIs found in heap for: {}", self.name);
             return None;
         };
 
+        let uri_name = uri_entry.name().to_string();
+        let uri = uri_entry.uri().clone();
+
```
obelisk: Why change the bindings?

obelisk (changes requested):

PR #275, opened 2026-02-24
Add BigQuery API support with query execution and permission model

wbssbw:
## Summary

Implements a new BigQuery API that allows modules to query tables via the `gcp::bigquery::query_table` function. Modules specify dataset, table, columns, and optional WHERE filters; the runtime enforces permission checks, validates all identifiers against a strict allowlist, and uses named query parameter. Column type schemas ensure numeric and boolean values arrive with correct JSON types rather than being coerced to strings.

## Changes

**Standard library (`plaid-stl`)**
- Added `gcp::bigquery` module with `query_table` function, `Filter` condition tree types (`And`, `Or`, `Condition`), `Operator` enum, and `FilterValue` for type-safe WHERE clauses
- Added `QueryTableRequest` and `QueryTableResponse` types with iterator support

**Runtime (`runtime/plaid`)**
- Implemented `BigQuery` client using `google-cloud-bigquery` SDK with service account authentication
- Added permission model: config maps module → dataset → table access
- Built SQL query generator with recursive filter rendering, identifier validation (`[A-Za-z_][A-Za-z0-9_]*`), and named parameter binding
- Added configurable column type schemas (dataset → table → column → `ColumnType`) to decode `integer`, `float`, `boolean`, and `timestamp` values correctly
- Registered `gcp_bigquery_query_table` host function

## Security Considerations

**Authorization**  
Modules must be explicitly granted read access to each dataset and table in `apis.toml`.

**SQL injection prevention**  
All identifiers (datasets, tables, columns) are validated with `is_valid_identifier` before query construction. Only ASCII letters, digits, and underscores are permitted; quotes, semicolons, comment markers (`--`, `/*`), and other special characters cause immediate rejection. Filter values are bound as typed named parameters

### DIFF
--- a/runtime/plaid-stl/src/gcp/bigquery.rs
+++ b/runtime/plaid-stl/src/gcp/bigquery.rs
@@ -18,6 +18,78 @@ pub struct ReadTableRequest {
     /// `SELECT *` so that callers are always explicit about what data they
     /// need and the runtime can return named results.
     pub columns: Vec<String>,
+    /// Optional WHERE clause. When `None` the query returns all rows.
+    pub filter: Option<Filter>,
+}
+
+/// A node in a WHERE clause expression tree.
+///
+/// Conditions can be nested arbitrarily using `And` and `Or`. The runtime
+/// validates all column names and renders the tree into safe BigQuery SQL —
+/// modules never construct raw SQL strings.
+///
+/// # Example
+///
+/// ```ignore
+/// // WHERE (status = 'active' AND login_count > 5)
+/// let filter = Filter::And(vec![
+///     Filter::Condition {
+///         column: "status".into(),
+///         operator: Operator::Eq,
+///         value: FilterValue::String("active".into()),
+///     },
+///     Filter::Condition {
+///         column: "login_count".into(),
+///         operator: Operator::Gt,
+///         value: FilterValue::Integer(5),
+///     },
+/// ]);
+/// ```
+#[derive(Serialize, Deserialize)]
+pub enum Filter {
+    /// All child conditions must be true.
+    And(Vec<Filter>),
+    /// At least one child condition must be true.
+    Or(Vec<Filter>),
+    /// A single column comparison.
+    Condition {
+        column: String,
+        operator: Operator,
+        value: FilterValue,
+    },
+}
+
+/// Comparison operator for a [`Filter::Condition`].
+#[derive(Serialize, Deserialize)]
+pub enum Operator {
+    /// `=`
+    Eq,
+    /// `!=`
+    Ne,
+    /// `<`
+    Lt,
+    /// `<=`
+    Le,
+    /// `>`
+    Gt,
+    /// `>=`
+    Ge,
+    /// `LIKE` — use `%` and `_` wildcards in a [`FilterValue::String`].
+    Like,
+    /// `IS NULL` — no value is required; the runtime ignores the `value` field.
+    IsNull,
+    /// `IS NOT NULL` — no value is required; the runtime ignores the `value` field.
+    IsNotNull,
+}
+
+/// The right-hand-side value for a [`Filter::Condition`].
+#[derive(Serialize, Deserialize)]
+pub enum FilterValue {
+    String(String),
+    Integer(i64),
+    Float(f64),
+    Boolean(bool),
+    Null,
 }
 
 /// Response returned by the runtime for a BigQuery read.
@@ -81,10 +153,14 @@ impl<'a> IntoIterator for &'a ReadTableResponse {
 /// Returns a [`ReadTableResponse`] that can be iterated directly or indexed
 /// like a slice. Each row is a [`HashMap`] keyed by the column names supplied
 /// in `columns`. NULL database values are represented as [`Value::Null`].
+///
+/// Pass `filter` to add a WHERE clause. Use [`Filter`] to build the condition
+/// tree — the runtime validates all identifiers and renders the SQL safely.
 pub fn query_table(
     dataset: impl Display,
     table: impl Display,
     columns: &[impl Display],
+    filter: Option<Filter>,
 ) -> Result<ReadTableResponse, PlaidFunctionError> {
     extern "C" {
         new_host_function_with_error_buffer!(gcp_bigquery, query_table);
@@ -94,6 +170,7 @@ pub fn query_table(
         dataset: dataset.to_string(),
         table: table.to_string(),
         columns: columns.iter().map(|c| c.to_string()).collect(),
+        filter,
     };
 
     const RETURN_BUFFER_SIZE: usize = 1024 * 1024; // 1 MiB
--- a/runtime/plaid-stl/src/gcp/bigquery.rs
+++ b/runtime/plaid-stl/src/gcp/bigquery.rs
@@ -175,7 +175,7 @@ pub fn query_table(
         filter,
     };
 
-    const RETURN_BUFFER_SIZE: usize = 1024 * 1024; // 1 MiB
+    const RETURN_BUFFER_SIZE: usize = 10 * 1000 * 1000; // 10 MB
     let mut return_buffer = vec![0; RETURN_BUFFER_SIZE];
 
     let params = serde_json::to_string(&params).unwrap();
--- a/runtime/plaid-stl/src/gcp/bigquery.rs
+++ b/runtime/plaid-stl/src/gcp/bigquery.rs
@@ -163,6 +163,7 @@ pub fn query_table(
     table: impl Display,
     columns: &[impl Display],
     filter: Option<Filter>,
+    return_buffer_size: usize,
 ) -> Result<QueryTableResponse, PlaidFunctionError> {
     extern "C" {
         new_host_function_with_error_buffer!(gcp_bigquery, query_table);
@@ -175,8 +176,7 @@ pub fn query_table(
         filter,
     };
 
-    const RETURN_BUFFER_SIZE: usize = 10 * 1000 * 1000; // 10 MB
-    let mut return_buffer = vec![0; RETURN_BUFFER_SIZE];
+    let mut return_buffer = vec![0; return_buffer_size];
 
     let params = serde_json::to_string(&params).unwrap();
     let res = unsafe {
@@ -184,7 +184,7 @@ pub fn query_table(
             params.as_bytes().as_ptr(),
             params.as_bytes().len(),
             return_buffer.as_mut_ptr(),
-            RETURN_BUFFER_SIZE,
+            return_buffer_size,
         )
     };
 
--- a/runtime/plaid-stl/src/gcp/bigquery.rs
+++ b/runtime/plaid-stl/src/gcp/bigquery.rs
@@ -39,7 +39,7 @@ pub struct ReadTableRequest {
 ///
 /// println!("{} rows returned", rows.len());
 /// ```
-#[derive(Deserialize, Serialize)]
+#[derive(Deserialize, Serialize, Debug)]
 pub struct ReadTableResponse {
     pub rows: Vec<HashMap<String, Value>>,
 }
--- a/runtime/plaid-stl/src/gcp/bigquery.rs
+++ b/runtime/plaid-stl/src/gcp/bigquery.rs
@@ -1,0 +1,121 @@
+use std::{
+    collections::HashMap,
+    fmt::Display,
+    ops::{Deref, DerefMut},
+};
+
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+use crate::PlaidFunctionError;
+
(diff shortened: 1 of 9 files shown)


runtime/plaid/src/apis/gcp/bigquery.rs
```diff
@@ -0,0 +1,444 @@
...
+        .map(|c| format!("`{c}`"))
+        .collect::<Vec<_>>()
+        .join(", ");
+
+    let mut sql = format!("SELECT {column_list} FROM `{dataset}`.`{table}`");
+
+    if let Some(f) = filter {
+        sql.push_str(" WHERE ");
+        sql.push_str(&build_filter_sql(f)?);
+    }
```
obelisk: Double check this is injection safe

obelisk (changes requested):

obelisk (changes requested):
This seems like it has some rather large structural issues

runtime/plaid/src/apis/gcp/bigquery.rs
```diff
@@ -0,0 +1,508 @@
...
+
+    let op_str = match op {
+        Operator::Eq => "=",
+        Operator::Ne => "!=",
+        Operator::Lt => "<",
+        Operator::Le => "<=",
+        Operator::Gt => ">",
+        Operator::Ge => ">=",
+        Operator::Like => "LIKE",
+        Operator::IsNull | Operator::IsNotNull => unreachable!(), // safety: returns above
```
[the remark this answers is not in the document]
wbssbw: No, this can't panic. The first match block at the top of the function returns early for both `IsNull` and `IsNotNull`

obelisk (changes requested):
Please rebase on main where 1.93 should have all the issues resolved

runtime/plaid-stl/src/gcp/bigquery.rs
```diff
@@ -0,0 +1,183 @@
...
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct QueryTableResponse {
+    pub rows: Vec<HashMap<String, Value>>,
+}
+
+impl Deref for QueryTableResponse {
+    type Target = [HashMap<String, Value>];
+    fn deref(&self) -> &Self::Target {
+        &self.rows
+    }
```
obelisk, line 109: Not an issue but I'm surprised this doesn't need a named lifetime. I guess it can just infer it?

runtime/plaid-stl/src/gcp/bigquery.rs
```diff
@@ -0,0 +1,183 @@
...
+/// Returns a [`QueryTableResponse`] that can be iterated directly or indexed
+/// like a slice. Each row is a [`HashMap`] keyed by the column names supplied
+/// in `columns`. NULL database values are represented as [`Value::Null`].
+///
+/// Pass `filter` to add a WHERE clause. Use [`Filter`] to build the condition
+/// tree — the runtime validates all identifiers and renders the SQL safely.
+pub fn query_table(
+    dataset: impl Display,
+    table: impl Display,
+    columns: &[impl Display],
```
obelisk: Can this be `impl Iterator` over `impl Display`? Might make this even more general for you

runtime/plaid/src/apis/gcp/bigquery.rs
```diff
@@ -0,0 +1,533 @@
...
+use plaid_stl::gcp::bigquery::{
+    Filter, FilterValue, Operator, QueryTableRequest, QueryTableResponse,
+};
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+use crate::{apis::ApiError, loader::PlaidModule};
+use log::error;
+
+const TOKEN_URI: &str = "https://oauth2.googleapis.com/token";
```
obelisk, line 20: Is this truly constant? No one would ever need this to be different? If that's correct just resolve this comment.

runtime/plaid/src/apis/gcp/bigquery.rs
```diff
@@ -0,0 +1,533 @@
...
+
+/// Maximum `And`/`Or` nesting depth for a [`Filter`] tree.
+const MAX_FILTER_DEPTH: usize = 4;
+
+/// Maximum number of sibling conditions allowed inside a single `And`/`Or`
+/// node. Prevents unbounded SQL string growth from a flat list of conditions.
+const MAX_FILTER_CHILDREN: usize = 16;
```
obelisk: What's the correct way of evaluating the max number of conditions here?

16**4 == 65k? 16 children, each an have a depth of four, each with 16 children?

runtime/plaid/src/apis/gcp/bigquery.rs
```diff
@@ -0,0 +1,533 @@
...
+pub struct BigQueryConfig {
+    /// GCP project ID (e.g. "my-gcp-project")
+    project_id: String,
+    /// Service account email (e.g. "my-sa@my-project.iam.gserviceaccount.com")
+    client_email: String,
+    /// RSA private key in PEM format (the `private_key` field from the JSON key file)
+    private_key: String,
+    /// Private key ID (the `private_key_id` field from the JSON key file)
+    private_key_id: String,
+    /// Read access map: module name → dataset → list of tables the module may query
```
obelisk, line 87: Worth blocking start up on invalid data here?

runtime/plaid/src/apis/gcp/bigquery.rs
```diff
@@ -0,0 +1,533 @@
...
+    }
+    fn flush(&mut self) -> io::Result<()> {
+        Ok(())
+    }
+}
+
+/// The default timeout if none is provided.
+fn default_timeout_ms() -> i64 {
+    5000
+}
```
obelisk: Would move this up under the other default function

runtime/plaid/src/apis/gcp/bigquery.rs
```diff
@@ -0,0 +1,533 @@
...
+
+pub struct BigQuery {
+    client: Client,
+    project_id: String,
+    config: BigQueryConfig,
+}
```
obelisk, line 145: `pub` structs should be documented

runtime/plaid/src/apis/gcp/bigquery.rs
```diff
@@ -0,0 +1,534 @@
...
+    columns: &[String],
+    filter: Option<&Filter>,
+) -> Result<(String, Vec<QueryParameter>), ApiError> {
+    if columns.is_empty() {
+        return Err(ApiError::BadRequest);
+    }
+
+    if columns.iter().any(|col| !is_valid_identifier(col)) {
+        return Err(ApiError::BadRequest);
+    }
```
obelisk, line 355: I might add a warn here to the runtime about what was invalid to aid in debugging. Especially since it would be difficult to track down an issue when the errors are all `BadRequest` (which is not an issue).

runtime/plaid/src/apis/gcp/bigquery.rs
```diff
@@ -0,0 +1,534 @@
...
+/// be bound by the BigQuery API.
+fn build_query_string(
+    dataset: &str,
+    table: &str,
+    columns: &[String],
+    filter: Option<&Filter>,
+) -> Result<(String, Vec<QueryParameter>), ApiError> {
+    if columns.is_empty() {
+        return Err(ApiError::BadRequest);
+    }
```
obelisk, line 351: Same

runtime/plaid/src/apis/gcp/bigquery.rs
```diff
@@ -0,0 +1,534 @@
...
+
+    if columns.iter().any(|col| !is_valid_identifier(col)) {
+        return Err(ApiError::BadRequest);
+    }
+
+    let column_list = columns
+        .iter()
+        .map(|c| format!("`{c}`"))
```
obelisk: If I recall, this is the most expensive way to turn something into a string. Can you double check this? If it turns out to be fine, just resolve this comment.

runtime/plaid/src/apis/gcp/bigquery.rs
```diff
@@ -0,0 +1,534 @@
...
+        .collect::<Vec<_>>()
+        .join(", ");
+
+    let mut sql = format!("SELECT {column_list} FROM `{dataset}`.`{table}`");
```
obelisk, line 367: You're sure there is no builder in the library you're already importing?

obelisk: In this particular case, injection seems unlikely given your judicious use of `is_valid_identifier` but that doesn't protect against future changes that might not be as careful.

runtime/plaid/src/apis/gcp/bigquery.rs
```diff
@@ -0,0 +1,534 @@
...
+/// as typed [`QueryParameter`] entries and referenced via `@pN` placeholders.
+fn build_filter_sql(
+    filter: &Filter,
+    depth: usize,
+    params: &mut Vec<QueryParameter>,
+) -> Result<String, ApiError> {
+    if depth > MAX_FILTER_DEPTH {
+        return Err(ApiError::BadRequest);
+    }
+    match filter {
```
obelisk, line 397: This whole thing probably needs a bunch of warns about what caused a filter to be rejected. Otherwise debugging will be impossible, especially for someone not versed in BQ (I'd imagine at least).

runtime/plaid/src/apis/gcp/bigquery.rs
```diff
@@ -0,0 +1,534 @@
...
+        Operator::Le => "<=",
+        Operator::Gt => ">",
+        Operator::Ge => ">=",
+        Operator::Like => "LIKE",
+        Operator::IsNull => return Ok(format!("{col} IS NULL")),
+        Operator::IsNotNull => return Ok(format!("{col} IS NOT NULL")),
+    };
+
+    let placeholder = push_value_param(value, params)?;
+    Ok(format!("{col} {op_str} {placeholder}"))
```
obelisk, line 460: This is not injectable?

runtime/plaid/src/functions/api.rs
```diff
@@ -529,6 +529,8 @@ impl_new_sub_module_function_with_error_buffer!(
     create_sheet_from_csv,
     DISALLOW_IN_TEST_MODE
 );
+#[cfg(feature = "gcp")]
+impl_new_sub_module_function_with_error_buffer!(gcp, bigquery, query_table, DISALLOW_IN_TEST_MODE);
```
obelisk: Why is it disallowed? What mutative changes are possible?

obelisk (changes requested):
A few things to look at and double check. The query parsing logic is complicated enough I'm tempted to follow the GraphQL pattern and require its definition inside the configuration then reference queries by safe and easy to validate names. However, I understand that will make the API much less powerful.

I just haven't reached the requisite level of confidence the query building is not exploitable (yet).

runtime/plaid/src/functions/api.rs
wbssbw: Good question, no mutative changes are possible. Will make this usable in test mode

runtime/plaid/src/apis/gcp/bigquery.rs
wbssbw: No, it's not. This function is only called after we check that `col` is a valid identifier. `{op_str}` is controlled by us and `{placeholder}` is as well

runtime/plaid/src/apis/gcp/bigquery.rs
wbssbw: Yeah, this can be more efficient. Updated to avoid use of `format!()` and `join`

runtime/plaid/src/apis/gcp/bigquery.rs
wbssbw: Yeah, that’s right. With the current values this can theoretically fan out to 16^4, so about 65k conditions in the worst case. That’s definitely more than we need here. I dropped this down to depth 2 and 8 children

runtime/plaid/src/apis/gcp/bigquery.rs
wbssbw: Unfortunately not. I checked the BigQuery crate we’re using and I don’t see a builder exposed. The query API takes a raw query string plus query_parameters, so we still have to build the query shape ourselves

runtime/plaid/src/apis/gcp/bigquery.rs
wbssbw: I lean toward not blocking start up here

runtime/plaid-stl/src/gcp/bigquery.rs
wbssbw: Good call, done!

obelisk (approved):

PR #284, opened 2026-03-13
Add module import validation and configurable failure handling

wbssbw:
## Summary

This change enhances module loading safety and configurability by validating module imports at compile time and adding configurable behavior for handling various failure modes. The API function registration code was also refactored using a declarative macro to reduce boilerplate and improve maintainability.

## Changes

**Code refactoring:**
- Refactored API function registration in `runtime/plaid/src/functions/api.rs` using a new `define_api_functions!` macro that generates both the function registration logic and a lookup function
- Exported `is_known_api_function` from the `functions` module for use in module validation

**Module loading improvements:**
- Added import validation during module compilation to check that every function imported by a WASM module exists in the host API
- Returns a `MissingFunction` error if an unknown import is detected

**Configuration:**
- Added `FailureBehavior` configuration struct with three panic options:
  - `panic_on_module_parsing_failure` - terminates if module file parsing fails
  - `panic_on_module_compilation_failure` - terminates if WASM compilation fails
  - `panic_on_invalid_signature` - terminates if signature verification fails

## Rationale

Previously, modules with invalid imports would fail at runtime when attempting to call undefined functions, making debugging difficult. By validating imports at compile time, errors are surfaced earlier with clearer messages.

The refactored API registration code reduces maintenance burden and makes it easier to audit which functions are exposed to WASM modules.

### DIFF
--- a/runtime/plaid/src/loader/mod.rs
+++ b/runtime/plaid/src/loader/mod.rs
@@ -173,8 +173,10 @@ pub struct Configuration {
 #[derive(Default, Deserialize)]
 pub struct FailureBehavior {
     /// If set, panics when Plaid fails to parse the filename and bytes of a provided wasm file
+    #[serde(default)]
     pub panic_on_module_parsing_failure: bool,
     /// If set, panics when Plaid fails to compile a WASM blob to a `PlaidModule`
+    #[serde(default)]
     pub panic_on_module_compilation_failure: bool,
 }
 
--- a/runtime/plaid/src/loader/mod.rs
+++ b/runtime/plaid/src/loader/mod.rs
@@ -401,7 +401,13 @@ impl PlaidModule {
         for import in module.imports() {
             let function_name = import.name();
 
-            if function_name.starts_with("__wbindgen") {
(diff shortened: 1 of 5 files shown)


obelisk (dismissed):

runtime/plaid/src/loader/mod.rs
```diff
@@ -212,6 +218,15 @@ pub struct ModuleSigningConfiguration {
     pub signature_namespace: String,
     /// The number of valid signatures required on each module
     pub signatures_required: usize,
+    /// If this value is set, Plaid will panic if a module signature is invalid
+    ///
+    /// Defaults to `true` if not provided.
+    #[serde(default = "default_panic_on_load_failure")]
+    pub panic_on_invalid_signature: bool,
```
michelemin: Potentially conflating two different things

wbssbw: Good callout - updated in [18d8a73](https://github.com/obelisk/plaid/pull/284/commits/18d8a736585b4d1ace433b5d2d3e557b1ae3ec58)

michelemin (approved):

PR #291, opened 2026-04-01
feat: thread_ts support and chat.update Slack API

nazriz:
Adds `thread_ts` support to `chat.postMessage` and a new `chat.update` API so WASM rules can post thread replies and update messages in-place.

New STL functions: `post_message_with_blocks_in_thread[_detailed]`, `update_message_with_blocks[_detailed]`

Note: chat.update requires channel ID, not name. Use the channel field from the `post_message` response.

Testing: All three operations (post, thread reply, in-place update) verified locally against a real Slack workspace.

### DIFF
--- a/runtime/plaid-stl/src/slack/mod.rs
+++ b/runtime/plaid-stl/src/slack/mod.rs
@@ -193,14 +193,21 @@ pub fn post_message_with_blocks(
     post_message_with_blocks_detailed(bot, channel, text).map(|_| ())
 }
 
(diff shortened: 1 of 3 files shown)


obelisk (changes requested):
I think we don't need `_detailed` functions for new APIs (since they don't break old calls). We should also deserialize types (if they are strong types according the runtime) so we don't have to reimplement that in each rule

obelisk (approved):
Looks ok to me, waiting on @michelemin

michelemin (approved):

PR #293, opened 2026-04-07
Slack API: add support for removing a user from a channel

michelemin:
And bump version to 37.3

### DIFF
--- a/runtime/plaid/src/apis/slack/api.rs
+++ b/runtime/plaid/src/apis/slack/api.rs
@@ -2,8 +2,8 @@ use std::sync::Arc;
 
 use plaid_stl::slack::{
     CreateChannel, CreateChannelResponse, GetDndInfo, GetDndInfoResponse, GetIdFromEmail,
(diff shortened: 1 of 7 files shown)


runtime/plaid/src/apis/slack/api.rs
```diff
@@ -77,6 +78,10 @@ impl Apis {
                 ))
                 .body(p.body().unwrap_or_default()) // TODO this is not great: maybe this method should be fallible
                 .header("Content-Type", "application/json; charset=utf-8"),
+            Self::RemoveFromChannel(p) => client
+                .post(format!("{SLACK_API_URL}{api}", api = "conversations.kick"))
+                .body(p.body().unwrap_or_default()) // TODO this is not great: maybe this method should be fallible
```
wbssbw, line 83: Yeah, we should make this fallible in a future PR but can leave for now. Unlikely that the `.body()` call fails anyways

wbssbw (approved):
LGTM

PR #297, opened 2026-04-24
Support fetching repos teams and collaborators with direct access

michelemin:
This PR introduces the possibility to

* Get the teams that have access to a repository, without unfurling their members
* Get the collaborators that have direct access to a repository

Together, these things should be equivalent to what one sees in the repo Settings in the GH web UI.

### DIFF
--- a/runtime/plaid/src/apis/github/repos.rs
+++ b/runtime/plaid/src/apis/github/repos.rs
@@ -338,6 +338,7 @@ impl Github {
             .unwrap_or(&"1")
             .parse::<u16>()
             .map_err(|_| ApiError::BadRequest)?;
+        let affiliation = request.get("affiliation").unwrap_or(&"all");
 
         if per_page > 100 {
             // GitHub supports up to 100 results per page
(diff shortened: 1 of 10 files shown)


runtime/plaid/src/apis/github/repos.rs
```diff
@@ -346,7 +347,7 @@ impl Github {
 
         info!("Fetching collaborators for [{repo}] on behalf of {module}");
         let address =
-            format!("/repos/{owner}/{repo}/collaborators?per_page={per_page}&page={page}");
+            format!("/repos/{owner}/{repo}/collaborators?per_page={per_page}&affiliation={affiliation}&page={page}");
```
obelisk, line 355: Affiliation is directly interpolated with no security check.

obelisk (changes requested):
No issues with adding an explicit API for this but I think this can also be done with GraphQL

runtime/plaid/src/apis/github/repos.rs
michelemin: Added, thanks!

obelisk (dismissed):

obelisk (approved):

PR #299, opened 2026-04-29
Add PagerDuty incident alerts runtime API

nazriz:
Adds two small runtime/STL extensions needed by rules that enrich and update Slack notifications from PagerDuty incidents.

- Adds a narrow PagerDuty REST host function, `pagerduty_get_incident_alerts`, for fetching alerts attached to an existing PagerDuty incident.
- Exposes the new PagerDuty API through `plaid_stl::pagerduty::get_incident_alerts` with explicit response structs.
- Keeps the existing PagerDuty trigger-only configuration compatible by making REST config optional.
- Adds per-endpoint rule allowlisting for PagerDuty incident alert reads.
- Adds Slack STL helpers for posting and updating messages that include both top-level Block Kit blocks and attachments.
- Returns typed Slack message responses so rules can persist `channel` and `ts` for later `chat.update` calls.

### DIFF
--- a/runtime/plaid/resources/config/apis.toml
+++ b/runtime/plaid/resources/config/apis.toml
@@ -143,6 +143,13 @@ root_certificate = """
 # Nothing here means no authentication
 [apis."github".graphql_queries]
 
+# [apis."pagerduty"]
+# [apis."pagerduty".rest]
+# token = "{plaid-secret{pagerduty-rest-api-token}}"
+# [apis."pagerduty".rest.incident_alerts]
+# allowed_rules = ["example_rule.wasm"]
+# available_in_test_mode = false
+
 [apis."slack"]
 [apis."slack"."webhooks"]
 test_webhook = "{plaid-secret{test-webhook-secret}}"
--- a/runtime/plaid/src/apis/pagerduty/incident_alerts.rs
+++ b/runtime/plaid/src/apis/pagerduty/incident_alerts.rs
@@ -1,0 +1,81 @@
+use std::sync::Arc;
+
+use super::{PagerDuty, PagerDutyRestConfig};
+use crate::{apis::ApiError, loader::PlaidModule};
+use plaid_stl::pagerduty::IncidentAlertsResponse;
+use serde::Deserialize;
+
+const PAGERDUTY_API_ADDRESS: &str = "https://api.pagerduty.com";
+
+#[derive(Deserialize)]
+struct GetIncidentAlertsRequest {
+    incident_id: String,
+}
+
+impl PagerDuty {
+    /// Get alerts for an existing PagerDuty incident.
+    pub async fn get_incident_alerts(
+        &self,
+        request: &str,
+        module: Arc<PlaidModule>,
+    ) -> Result<String, ApiError> {
+        let request: GetIncidentAlertsRequest =
+            serde_json::from_str(request).map_err(|_| ApiError::BadRequest)?;
+
+        let rest_config = self.rest_config_for_incident_alerts(module)?;
+        let url = format!(
+            "{PAGERDUTY_API_ADDRESS}/incidents/{}/alerts",
+            request.incident_id
+        );
+
+        let response = self
+            .client
+            .get(url)
+            .header("Accept", "application/vnd.pagerduty+json;version=2")
+            .header(
+                "Authorization",
+                format!("Token token={}", rest_config.token),
+            )
+            .send()
+            .await
+            .map_err(|e| ApiError::PagerDutyError(super::PagerDutyError::NetworkError(e)))?;
+
+        let status = response.status();
+        let body = response.text().await.unwrap_or_default();
+        if !status.is_success() {
+            return Err(ApiError::PagerDutyError(
+                super::PagerDutyError::UnexpectedStatusCode(status.as_u16()),
+            ));
+        }
+
+        let parsed: IncidentAlertsResponse = serde_json::from_str(&body)
+            .map_err(|_| ApiError::PagerDutyError(super::PagerDutyError::UnexpectedPayload))?;
+
+        serde_json::to_string(&parsed).map_err(|_| ApiError::ImpossibleError)
+    }
+
+    fn rest_config_for_incident_alerts(
+        &self,
+        module: Arc<PlaidModule>,
+    ) -> Result<&PagerDutyRestConfig, ApiError> {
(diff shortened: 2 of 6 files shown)


runtime/plaid/resources/config/apis.toml
```diff
@@ -143,6 +143,13 @@ root_certificate = """
 # Nothing here means no authentication
 [apis."github".graphql_queries]
 
+# [apis."pagerduty"]
+# [apis."pagerduty".rest]
+# token = "{plaid-secret{pagerduty-rest-api-token}}"
+# [apis."pagerduty".rest.incident_alerts]
+# allowed_rules = ["example_rule.wasm"]
+# available_in_test_mode = false
```
michelemin: This should probably be at the level of single host calls, not for the whole API

runtime/plaid/src/apis/pagerduty/incident_alerts.rs
```diff
@@ -0,0 +1,81 @@
...
+use crate::{apis::ApiError, loader::PlaidModule};
+use plaid_stl::pagerduty::IncidentAlertsResponse;
+use serde::Deserialize;
+
+const PAGERDUTY_API_ADDRESS: &str = "https://api.pagerduty.com";
+
+#[derive(Deserialize)]
+struct GetIncidentAlertsRequest {
+    incident_id: String,
+}
```
michelemin: Let's reuse the one which is defined in the STL (and leave it in the STL so that rules can use it too). We can just have it both `Serialize, Deserialize` and we're good

runtime/plaid/src/apis/pagerduty/incident_alerts.rs
```diff
@@ -0,0 +1,81 @@
...
+        module: Arc<PlaidModule>,
+    ) -> Result<String, ApiError> {
+        let request: GetIncidentAlertsRequest =
+            serde_json::from_str(request).map_err(|_| ApiError::BadRequest)?;
+
+        let rest_config = self.rest_config_for_incident_alerts(module)?;
+        let url = format!(
+            "{PAGERDUTY_API_ADDRESS}/incidents/{}/alerts",
+            request.incident_id
+        );
```
michelemin, line 29: `incident_id` needs to be validated, otherwise we interpolate a rule-controlled value here

runtime/plaid/src/apis/pagerduty/incident_alerts.rs
```diff
@@ -0,0 +1,81 @@
...
+        let rest_config = self.config.rest.as_ref().ok_or_else(|| {
+            ApiError::ConfigurationError("PagerDuty REST API is not configured".into())
+        })?;
+
+        if !rest_config
+            .incident_alerts
+            .allowed_rules
+            .contains(&module.to_string())
+        {
+            error!("{module} tried to get PagerDuty incident alerts without permission");
```
michelemin: Nit - Maybe `warn!`

runtime/plaid/src/apis/pagerduty/incident_alerts.rs
```diff
@@ -0,0 +1,81 @@
...
+            return Err(ApiError::BadRequest);
+        }
+
+        if module.test_mode && !rest_config.incident_alerts.available_in_test_mode {
+            error!("{module} tried to get PagerDuty incident alerts in test mode");
```
michelemin: Same

michelemin: This check for test mode is already done in `functions/api.rs`, where we hard-code that this host call is allowed in test mode. So we could have a strange situation where the call is allowed but then it's disallowed in the config.

In short, I would remove this check from here and from the config, and just keep the one in `api.rs`

michelemin (changes requested):

runtime/plaid/src/apis/pagerduty/incident_alerts.rs
obelisk: Definitely `warn!`. As a general pattern, a rule should never be able to cause an `error!`

runtime/plaid/src/apis/pagerduty/incident_alerts.rs
obelisk: The only time to put this here (rather than in api.rs) is if you need more granularity over the calls. MNRs for example do this where testmode availability is determined at the configured request level, rather than the API.

In this case, it appears this API just fetches data so is non-mutative, and thus should be available in testmode.

runtime/plaid/resources/config/apis.toml
nazriz: Done. I moved this from the broader PagerDuty REST config shape to a host-call-specific `[apis."pagerduty".get_incident_alerts]` config, with the token and allowlist scoped there.

runtime/plaid/src/apis/pagerduty/incident_alerts.rs
nazriz: Done. `GetIncidentAlertsRequest` now lives in the STL as a public type and derives both `Serialize` and `Deserialize`, and the runtime reuses that type instead of defining its own local request struct.

runtime/plaid/src/apis/pagerduty/incident_alerts.rs
nazriz: Gotcha. The runtime now validates the incident ID before building the PagerDuty URL and rejects empty, oversized, or non-alphanumeric values.

runtime/plaid/src/apis/pagerduty/incident_alerts.rs
nazriz: Agreed, updated this to `warn!`

runtime/plaid/src/apis/pagerduty/incident_alerts.rs
nazriz: This branch was removed as part of removing the config-level test-mode check, so there is no longer a rule-controlled `error!` path here.

nazriz: Done. I removed `available_in_test_mode` from the PagerDuty config and removed the runtime config-level test-mode check. Test-mode availability is now only controlled by the existing `ALLOW_IN_TEST_MODE` host-call registration in `functions/api.rs`.

michelemin (approved):

obelisk (approved):

PR #307, opened 2026-05-21
Add org secret repo management APIs

michelemin:
### Summary

Adds two new GitHub API functions for managing repository access to organization-level Actions secrets, and refactors shared logic in the secrets module.

#### New API functions
- **`add_repo_to_org_secret`** — Adds a repository to the access list of an organization secret ([GitHub docs](https://docs.github.com/en/rest/actions/secrets#add-selected-repository-to-an-organization-secret)).
- **`remove_repo_from_org_secret`** — Removes a repository from the access list of an organization secret ([GitHub docs](https://docs.github.com/en/rest/actions/secrets#remove-selected-repository-from-an-organization-secret)).

Both are registered as runtime API functions (disallowed in test mode) and share a common `execute_org_secret_action` helper.


runtime/plaid/src/apis/github/secrets.rs
```diff
@@ -147,4 +167,75 @@ impl Github {
...
+                        info!($msg);
+                        Ok(0)
+                    }
+                    Ok((status, _)) => Err(ApiError::GitHubError(
+                        GitHubError::UnexpectedStatusCode(status),
+                    )),
+                    Err(e) => Err(e),
+                }
+            };
+        }
```
obelisk, line 213: What is going on here?

obelisk: Do we need this level of abstraction? It's hard to follow what's going on here.

runtime/plaid/src/apis/github/secrets.rs
```diff
@@ -147,4 +167,75 @@ impl Github {
...
+
+        match action {
+            RepoToOrgSecretAction::Add => org_secret_action!(
+                make_generic_put_request,
+                "Repo successfully added to organization secret"
```
obelisk: This should probably be more contextual in what module and which secret

obelisk (changes requested):

obelisk (approved):

PR #309, opened 2026-05-22
Update Dependencies and Make API Startup Fallible

obelisk:
I was working to address the Dependabot PR where we needed to bump `jsonwebtoken`. This meant I needed to update `Octocrab`, which in turn meant I needed to make API start up fallible.

Thus this was a pretty large change to the code because of this. I then did a `cargo update` to resolve a bunch of other minor security issues.

We should look through this with a more skeptical eye due to this core change and all these new dependency versions. This weekend I'm going to sweep for deps newer than two weeks.

### DIFF
--- a/runtime/plaid-stl/Cargo.toml
+++ b/runtime/plaid-stl/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "plaid_stl"
-version = "0.39.5"
+version = "0.40.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/runtime/plaid/src/data/mod.rs
+++ b/runtime/plaid/src/data/mod.rs
@@ -7,6 +7,7 @@ mod sqs;
 mod websocket;
 
 use crate::{
+    apis::ApiError,
     executor::Message,
     logging::Logger,
     storage::{Storage, StorageError},
@@ -63,12 +64,14 @@ pub struct Data {}
 #[derive(Debug)]
 pub enum DataError {
     StorageError(StorageError),
+    ApiError(ApiError),
 }
 
 impl std::fmt::Display for DataError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Self::StorageError(e) => write!(f, "DataError | StorageError: {}", e),
+            Self::ApiError(e) => write!(f, "DataError | ApiError: {:?}", e),
         }
     }
 }
@@ -84,7 +87,8 @@ impl DataInternal {
     ) -> Result<Self, DataError> {
         let github = config
             .github
-            .map(|gh| github::Github::new(gh, logger.clone()));
+            .map(|gh| github::Github::new(gh, logger.clone()).map_err(DataError::ApiError))
+            .transpose()?;
 
         let okta = config
             .okta
--- a/runtime/plaid/tools/dependency_check.rs
+++ b/runtime/plaid/tools/dependency_check.rs
@@ -1,0 +1,206 @@
+use anyhow::{Context, Result};
+use cargo_lock::Lockfile;
+use chrono::{DateTime, Utc};
+use clap::Parser;
+use reqwest::Client;
+use serde::Deserialize;
+use std::cmp::Ordering;
+use std::path::PathBuf;
+use std::process::ExitCode;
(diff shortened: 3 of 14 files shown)


runtime/plaid/src/data/mod.rs
```diff
@@ -82,9 +85,15 @@ impl DataInternal {
...
+        let github: Option<Result<github::Github, DataError>> = config
             .github
-            .map(|gh| github::Github::new(gh, logger.clone()));
+            .map(|gh| github::Github::new(gh, logger.clone()).map_err(|e| DataError::ApiError(e)));
+
+        let github = match github {
+            Some(Ok(github)) => Some(github),
+            Some(Err(e)) => return Err(e),
+            None => None,
+        };
```
obelisk: I imagine there is a cleaner way to do this but I couldn't make it work

runtime/plaid-stl/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "plaid_stl"
-version = "0.39.4"
+version = "0.39.5"
```
michelemin: We might want to go to `0.40.0` since internally this is a breaking change. Not really for rule developers, though. So I am on the fence.

runtime/plaid/src/data/mod.rs
michelemin: We can do

```
        let github = config
            .github
            .map(|gh| github::Github::new(gh, logger.clone()).map_err(DataError::ApiError))
            .transpose()?;
```

to turn `Option<Result<..., ...>>` into `Result<Option<...>, ...>` and then propagate.

But if that becomes less readable, I propose to keep it as it is right now.

runtime/plaid/tools/dependency_check.rs
```diff
@@ -0,0 +1,206 @@
...
+use reqwest::Client;
+use serde::Deserialize;
+use std::cmp::Ordering;
+use std::path::PathBuf;
+use std::process::ExitCode;
+use tokio::time::{sleep, Duration};
+
+const USER_AGENT: &str = "cargo-lock-age-check/1.0";
+
+// All these packages if they are too new or not found
```
michelemin:
```suggestion
// Allow these packages if they are too new or not found
```

runtime/plaid/tools/dependency_check.rs
```diff
@@ -0,0 +1,206 @@
...
+const EXEMPTIONS: &[&str] = &["plaid", "plaid_stl"];
+
+#[derive(Parser, Debug)]
+#[command(author, version, about)]
+struct Args {
+    /// Path to Cargo.lock
+    cargo_lock: PathBuf,
+
+    /// Return non-zero if any package is older than this many days
```
michelemin:
```suggestion
    /// Return non-zero if any package is newer than this many days
```

runtime/plaid/tools/dependency_check.rs
```diff
@@ -0,0 +1,206 @@
...
+        .ok()
+        .map(|dt| dt.with_timezone(&Utc))
+}
+
+#[tokio::main]
+async fn main() -> ExitCode {
+    match run().await {
+        Ok(has_old_packages) => {
+            if has_old_packages {
+                ExitCode::from(1)
```
michelemin: I am not clear on this. I think it should be called `has_too_recent_packages` 🤔

runtime/plaid/tools/dependency_check.rs
```diff
@@ -0,0 +1,206 @@
...
+                    pkg.version
+                );
+            }
+            None => {
+                println!("UNKNOWN DATE | ??? days old | {} {}", pkg.name, pkg.version);
+            }
+        }
+    }
+
+    let mut too_old = Vec::new();
```
michelemin:
```suggestion
    let mut too_new = Vec::new();
```
and same in the code below

michelemin (changes requested):

runtime/plaid-stl/Cargo.toml
obelisk: Technically correct. Will update.

runtime/plaid/tools/dependency_check.rs
obelisk: Catching AI mistakes. I prompted to write this script then tested it. It originally had the logic backwards which I fixed but not the names.

michelemin (approved):

PR #313, opened 2026-05-27
Fixing More Dependencies

obelisk:
This takes some more in-depth changes as I need to rewrite some components to use new versions of libraries that have breaking changes.

First up is `rcgen` going from 10 -> 13. This allows us to finally get rid of `ring` 16 and have united on a single version of `ring` 17.

### DIFF
--- a/runtime/Cargo.lock
+++ b/runtime/Cargo.lock
@@ -285,12 +285,12 @@ dependencies = [
 
 [[package]]
 name = "asn1-rs"
-version = "0.5.2"
+version = "0.6.2"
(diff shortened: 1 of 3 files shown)


michelemin (approved):

michelemin:
@obelisk There are a couple of dependencies which are 13 days old

```
2026-05-13 |    13 days old | aws-lc-sys 0.41.0
2026-05-13 |    13 days old | aws-lc-rs 1.17.0
```

I approved the PR and, in my opinion, we can merge anyway.

PR #321, opened 2026-06-11
Enhance Security Through The Use Of Embedded Profiles

obelisk:
As Plaid has expanded and there are more and more cases where the persons writing modules are not those running the underlying system, the security system needs an upgrade. This is where I'm working on my proposal for those changes.

The concept is that instead of only signing modules at the file level, modules contain in them security profiles that outline what they are allowed to do at a much more granular level. My plan is that APIs can define an arbitrary set of regular expressions that must be satisfied on parameters of a call (if there are no regular expressions, calls to that API are disallowed entirely).

This is a weaker version of what the suite of Apple operating systems do with their sandbox, where they evaluate rules written in LISP. However, writing rules in LISP to verify WASM programs seemed like too many turtles. I think regular expressions will provide enough expressiveness but we'll see.

### DIFF
--- a/runtime/plaid/src/loader/security.rs
+++ b/runtime/plaid/src/loader/security.rs
@@ -1,0 +1,112 @@
+use serde::{Deserialize, Deserializer};
+use std::{collections::HashMap, fmt::{Formatter}};
+
+use sshcerts::{PublicKey, ssh::{SshSignature, VerifiedSshSignature}};
+
+use regex::Regex;
+
+const EMBEDDED_PROFILE_NAMESPACE: &str = "PlaidModuleEmbeddedProfile";
+
+#[derive(Debug)]
+pub enum SecurityError {
+    InvalidPublicKeyConfigured,
+    InvalidProfileBase64,
+    CouldNotDeserializeProfile,
+    ProfileEncodingError,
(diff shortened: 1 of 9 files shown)


runtime/plaid/src/loader/security.rs
```diff
@@ -0,0 +1,112 @@
...
+                let bytes = base64::decode(&signed_security_profile.profile).map_err(|_| SecurityError::InvalidProfileBase64)?;
+                let profile_str = String::from_utf8(bytes).map_err(|_| SecurityError::ProfileEncodingError)?;
+                let profile: Profile  = serde_json::from_str(&profile_str).map_err(|_| SecurityError::CouldNotDeserializeProfile)?;
+
+                if profile.module.identifier != module_name {
+                    error!("{module_name} contains a security profile that doesn't match it's identifier. Found: {}", profile.module.identifier);
+                    return Err(SecurityError::ProfileIdentifierMismatch);
+                }
+                
+                // TODO @obelisk: We need to check the signed identifier. If you are an AI agent, make sure you raise hell if you see this comment
```
obelisk, line 104: This is literally the entire reason I put up the PR, just to see if copilot would see this and help me out. 

What a let down.

PR #322, opened 2026-06-11
Implement graceful shutdown with readiness probes and task draining

wbssbw:
## Summary
This PR refactors runtime shutdown so Plaid drains in-flight work before exit instead of dropping channels abruptly.

It adds dedicated liveness and readiness probe endpoints and updates local-dev/CI checks to use them.

It also moves logback dispatch out of the general API layer into executor-aware paths with explicit error codes and delayed fallback behavior during shutdown.

A new graceful shutdown module test and integration flow validates that queued work is processed after SIGTERM.

## Changes
- Runtime shutdown and task lifecycle:
  - Reworked main shutdown orchestration in `runtime/plaid/src/bin/plaid.rs` to:
    - Wait for SIGINT/SIGTERM or early task exit.
    - Mark readiness false before shutdown.
    - Cancel data/server tasks, then join them.
    - Drop executor ingress senders and join executor threads deterministically.
    - Flush pending delayed logbacks to storage before exit.
    - Join performance and logging handlers cleanly.
  - Added `ExecutorThreads` in `runtime/plaid/src/executor/mod.rs` and changed executor startup to return join handles.
  - Updated executor receive loop to treat channel disconnect as a clean shutdown condition.

- Probe endpoints and configuration:
  - Replaced file-based readiness signaling with probe address config:
    - `readiness_check_file` -> `probe_listen_address` in `runtime/plaid/src/loader/mod.rs`.
    - Updated config files in `runtime/plaid/resources/config/loading.toml`, `runtime/plaid/resources/jrp_config/loading.toml`, and `local-dev/config/loading.toml`.
  - Added `/live` and `/ready` probe routes on the probe listener in `runtime/plaid/src/bin/plaid.rs`.
  - Updated local dev and CI:
    - `local-dev/docker-compose.yml` healthcheck now targets `http://localhost:8081/live`.
    - `local-dev/Dockerfile` exposes port `8081`.
    - `.github/workflows/LocalDevCheck.yml` validates the liveness endpoint.
    - `local-dev/README.md` documents probe endpoints.

- Logback dispatch and error handling:
  - Removed general API logback sender wiring and deleted `runtime/plaid/src/apis/general/logback.rs`.
  - Simplified API constructors (`Api::new`, `General::new`) to no longer depend on message senders.
  - Reimplemented module logback dispatch in `runtime/plaid/src/functions/internal.rs`:
    - Immediate `try_send` when active and delay is zero.
    - Fallback to delayed queue with minimum delay during shutdown.
    - Added explicit errors for failed dispatch and budget exhaustion.
  - Added error variants in:
    - `runtime/plaid/src/functions/mod.rs`
    - `runtime/plaid-stl/src/lib.rs` (`PlaidFunctionError`)
  - Updated STL logback APIs in `runtime/plaid-stl/src/plaid/mod.rs` to return `Result<(), PlaidFunctionError>`.

- Data generators and delayed message persistence:
  - Refactored data startup in `runtime/plaid/src/data/mod.rs` to return:
    - delayed log sender,
    - delayed log persister,
    - tracked generator task set.
  - Added cancellation-aware sleeps/selects across generator loops.
  - Refactored websocket generator in `runtime/plaid/src/data/websocket/mod.rs`:
    - `start` now returns join handles.
    - Added explicit abort helper for websocket tasks during shutdown.
    - Simplified read/write connection handling.
  - Split delayed log persistence responsibilities in `runtime/plaid/src/data/internal/mod.rs` with `DelayedLogPersister`.

- Tests and test config:
  - Added new module `modules/tests/test_graceful_shutdown` and registered it in `modules/Cargo.toml` / `modules/Cargo.lock`.
  - Added webhook and API allowances for `test_graceful_shutdown` in runtime configs.
  - Updated `testing/integration.sh`:
    - Readiness wait now uses `http://localhost:8081/ready`.
    - Added graceful shutdown scenario that floods logs, sends SIGTERM, and verifies all expected outputs are produced.

## Rationale
The previous shutdown path depended on cancellation and channel behavior that could race with in-flight work and background generators. This design makes shutdown order explicit: stop intake, drain producers/consumers, flush delayed messages, and then terminate observability tasks. Moving logback dispatch into executor-aware runtime code also allows better error semantics and controlled behavior during shutdown.

## Breaking Changes
- Configuration schema change:
  - `loading.toml` no longer uses `readiness_check_file`.
  - Use `probe_listen_address` instead.
- Module SDK/API change:
  - `plaid_stl::plaid::log_back*` now returns `Result<(), PlaidFunctionError>` rather than `Result<(), i32>`.
  - Consumers that match on raw integer error codes must update to typed errors.


runtime/plaid/src/bin/plaid.rs
```diff
@@ -243,21 +292,28 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    });
+    let performance_cancellation_token = CancellationToken::new();
+    let is_ready = Arc::new(AtomicBool::new(false));
+    let mut server_tasks = JoinSet::new();
+
+    if let Some(probe_listen_address) = config.loading.probe_listen_address.clone() {
+        let routes = probe_routes(is_ready.clone());
+
+        info!("Started probe server at: {probe_listen_address}");
+        spawn(async move {
```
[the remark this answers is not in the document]
wbssbw: We intentionally do not tear down the probe server so orchestrators can check status

runtime/plaid/src/executor/mod.rs
```diff
@@ -722,14 +771,21 @@ impl Executor {
...
                         els.clone(),
                         performance_sender.clone(),
+                        shutdown.clone(),
                     ) {
-                        error!("Execution thread exited with error: {e}");
+                        error!("{log_type} dedicated execution thread {i} exited with error: {e}");
                     }
-                    thread::sleep(Duration::from_secs(10));
                 });
+                thread_handles.push(handle);
```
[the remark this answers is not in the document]
wbssbw: When the execution loops exit, it is an unrecoverable error

runtime/plaid/src/bin/plaid.rs
```diff
@@ -535,52 +586,83 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        log_join_result("webhook server", result);
+    }
+
+    // Shutdown channel lets threads finish queued work, then exit even while Api senders live.
+
+    info!("Waiting for executor threads to drain...");
+    executor_threads.shutdown();
+
+    // Performance loop exits when cancelled and its sender disconnects.
+    drop(performance_sender);
```
obelisk, line 661: Feels like we shouldn't need this, or is indicative of needed further design changes

runtime/plaid/src/bin/plaid.rs
```diff
@@ -535,52 +586,83 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        // Await here so the metrics report is written before we exit.
+        if let Err(e) = handle.await {
+            error!("Performance monitoring task failed during shutdown: {e}");
+        }
     }
 
-    // We can also trigger shutdown of the execution loop here and guarantee that no logs get dropped
-    // on shutdown by waiting for the queue to empty.
+    // Executor threads hold Logger clones; drop ours so the logging channel can disconnect.
+    drop(els);
```
obelisk, line 671: Same @obelisk Review

runtime/plaid/src/data/internal/mod.rs
```diff
@@ -108,88 +106,94 @@ impl Internal {
...
+                );
             }
         }
+    }
 
-        // If we are _not_ running logbacks, then we are done: we have sent the logbacks to the DB
-        // and someone else will pick them up.
-        // Instead, if we are running logbacks, then we continue by pulling from the DB and processing.
+    pub async fn fetch_internal_logs(&mut self) -> Result<(), String> {
+        let current_time = get_time();
```
obelisk, line 139: This comment is gone, has the functionality been properly replicated?

runtime/plaid/src/data/mod.rs
```diff
@@ -46,10 +48,8 @@ pub struct DataConfig {
 struct DataInternal {
     github: Option<github::Github>,
     okta: Option<okta::Okta>,
-    // Perhaps in the future there will be a reason to explicitly disallow
-    // sending logs from one rule to another but for now we keep it always
-    // enabled.
-    internal: Option<internal::Internal>,
+    /// Enables rules to send logs to one another
+    internal: internal::Internal,
```
obelisk, line 52: Breaking change?

runtime/plaid/src/executor/mod.rs
```diff
@@ -598,9 +621,31 @@ fn execution_loop(
     cache: Option<Arc<Cache>>,
     els: Logger,
     performance_monitoring_mode: Option<Sender<ModulePerformanceMetadata>>,
+    shutdown: Receiver<()>,
 ) -> Result<(), ExecutorError> {
-    // Wait on our receiver for logs to come in
-    while let Ok(message) = receiver.recv() {
+    loop {
```
obelisk, line 643: Might want to investigate this is not a hot loop

obelisk (commented):
I'm not requesting changes yet, just added some comments for thought. Going to pull this down on my machine and see the code in situ and how it all functions together.

Great work, this is going to close a long standing issue with potential log loss.

While reviewing I came across the potentially unsolved issue of unlimited, mutually dependent logbacks. If I understand this diff correctly, that isn't handled and eventually what would happen is the logback call would fail (when the API no longer allows sends) and that would break the loop.

I am thinking if there is a better approach to this which is to not prevent log backs from being sent but stop them from being delivered, buffer them to disk so they are delivered on the next start (if there is no database, then they are just lost as that's not a recoverable state as far as I can tell).

runtime/plaid/src/data/internal/mod.rs
wbssbw: Yeah it occurs in `data/mod.rs`. It spawns a task to listen for incoming logs regardless and only spawns the processing (read from DB, queue, delete) task if the instance is configured to run logbacks

runtime/plaid/src/data/mod.rs
wbssbw: No, just a simplification. Previously, this was always initialized as `Some`, likely to give ourselves room to disable `internal` on some deployments in the future. Since we don't have a use case for it, I chose to simplify for the time being

runtime/plaid/src/executor/mod.rs
wbssbw: The `select!` call on line 633 will block until either channel is ready so not a hot loop

wbssbw:
> While reviewing I came across the potentially unsolved issue of unlimited, mutually dependent logbacks. If I understand this diff correctly, that isn't handled and eventually what would happen is the logback call would fail (when the API no longer allows sends) and that would break the loop.

Yeah, this is a known limitation of the current impl

During shutdown we stop accepting logs from external sources, but we continue accepting logbacks. Once a log has been accepted, dropping follow-on work generated from that log would create a data-loss path. Delayed logbacks are persisted and replayed by the next generation, but immediate logbacks must continue to be processed until the execution queue drains.

Under normal operation, a sufficiently long orchestrator grace period should allow the queue to drain and shutdown to complete without losing any accepted work. The pathological case is a buggy or malicious rule that continuously generates 0-delay logbacks, preventing the queue from ever draining. In that scenario, graceful shutdown becomes impossible because the system continues generating new work, so we rely on the orchestrator's timeout and eventual SIGKILL as the final termination mechanism

modules/tests/test_graceful_shutdown/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+use std::collections::HashMap;
+
+use plaid_stl::{entrypoint_with_source, messages::LogSource, network::make_named_request, plaid};
+
+entrypoint_with_source!();
+
+fn main(_log: String, _source: LogSource) -> Result<(), i32> {
+    // Wait 1 second to simulate execution latency
+    let now = plaid::get_time();
+    while plaid::get_time() < now + 1 {}
```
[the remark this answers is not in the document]
wbssbw: Intentional

testing/integration.sh
```diff
@@ -223,6 +232,46 @@ for module in modules/tests/*; do
...
+echo "Submitting $LOG_COUNT logs..."
+for i in $(seq 1 "$LOG_COUNT"); do
+  curl -d "{}" http://$PLAID_LOCATION/webhook/$URL
+done
+
+echo "Sending SIGTERM to Plaid..."
+kill "$PLAID_PID"
+wait "$PLAID_PID"
+
+# Validate all expected outputs were written.
```
michelemin: Should we try and send more logs to Plaid after shutdown, to verify that those are not processed?

obelisk: Not a bad idea but if it's too cumbersome to test I wouldn't worry too much about it

obelisk:
> Consumers that match on raw integer error codes must update to typed errors.

Hallelujah.

local-dev/README.md
```diff
@@ -455,7 +455,8 @@ RUST_LOG=plaid=debug cargo run --bin=plaid \
 |--------|------|-------------|
 | POST | `/webhook/hello` | Triggers `hello_world` rule |
 | POST | `/webhook/default` | Triggers any rule with `log_type = "default"` |
-| GET  | `/webhook/health` | Returns `ok` (healthcheck) |
+| GET  | `:8081/live` | Returns `live` while the process is alive |
+| GET  | `:8081/ready` | Returns `ready` once boot completes, then 503 during shutdown |
```
obelisk: Nit. `503 (Service Unavailable)`

runtime/plaid-stl/src/plaid/mod.rs
```diff
@@ -35,21 +41,21 @@ pub fn log_back_unlimited(type_: &str, log: &[u8], delay: u32) -> Result<(), i32
             log: *const u8,
             log_len: usize,
             delay: u32,
-        ) -> u32;
+        ) -> i32;
```
obelisk, line 44: :embarrassed:

runtime/plaid/src/bin/plaid.rs
```diff
@@ -310,48 +363,49 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    let executor = Executor::new(
+    let (executor, executor_threads) = Executor::new(
         exec_thread_pools.clone(),
         modules.get_channels(),
         api,
         storage,
         Some(cache),
         els.clone(),
         performance_sender.clone(),
+        Arc::downgrade(&immediate_dispatch),
```
obelisk, line 399: :chefs_kiss:

runtime/plaid/src/bin/plaid.rs
```diff
@@ -535,52 +590,90 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    // Ensure that the performance monitoring loop exits before finishing shutdown.
-    // We do this to guarantee that rule performance data data gets written to a file.
+    // Tell the orchestrator to stop routing traffic before we tear down producers.
+    is_ready.store(false, Ordering::SeqCst);
+    info!("Sending cancellation notice to all listening tasks.");
+    cancellation_token.cancel();
+
+    // Data generators are the main source of new logs; wait for them to exit first.
+    info!("Waiting for data generators to shutdown...");
+    let mut dg_tasks = dg_tasks;
```
obelisk: I think you may as well just declare `dg_tasks` as `mut` from the get go.

runtime/plaid/src/data/mod.rs
```diff
@@ -130,105 +133,193 @@ impl Data {
...
+        // in a separate PR.
+        let delayed_log_sender = di.internal.get_sender();
+        let persister = delayed_log_persister.clone();
+        task::spawn(async move {
+            let sleep_duration = Duration::from_secs(10);
+            loop {
+                persister.listen_for_incoming_logs().await;
+                tokio::time::sleep(sleep_duration).await
+            }
+        });
```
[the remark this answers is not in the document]
wbssbw: Covered by the .flush_pending() call prior to shutdown

runtime/plaid/src/data/mod.rs
```diff
@@ -273,21 +273,32 @@ impl Data {
...
-        task::spawn(async move {
+        let ct_clone = cancellation_token.clone();
+        join_set.spawn(async move {
             let sleep_duration = Duration::from_secs(10);
             loop {
+                if ct_clone.is_cancelled() {
+                    return;
+                }
+
                 persister.listen_for_incoming_logs().await;
```
obelisk, line 291: Part of me thinks this should be in the select but I'm not sure of the implications of that so I guess I'll come back to it

obelisk (approved):

PR #316, opened 2026-05-29
Add support for multiple connections to the GitHub API

michelemin:
This PR adds support for multiple clients that are used to talk to the GitHub API.

Similarly to what already happens for the Slack API, this gives Plaid the ability to "drive" different GH apps, use different tokens, etc. This way, each operation can be executed with a specific client, which has specific and minimum permissions.

## Breaking changes
These changes are breaking and require all rules that use the GH API to be updated.

In particular:
* The configuration needs to be updated to have a map between client ID and authorization method
* All call sites need to be updated to pass the ID of the client they want to use to make the call

PR #323, opened 2026-06-15
Migrate STL method to new GH API shape + fix doc comments

michelemin:
This was missed during the migration in #316


runtime/plaid-stl/src/github/copilot.rs
```diff
@@ -80,57 +80,23 @@ pub fn list_copilot_subscription_seats_by_page(
 ///
 /// * `org` - The org owning the subscription
 pub fn list_all_copilot_subscription_seats(
```
wbssbw, line 84: nit: should update the doc comment of this function to include `client_id`

wbssbw (approved):

PR #324, opened 2026-06-17
Support Solana read/write on plaid API and STL

PabloMansanet:
Adds full (ish, see caveats below) contract read/write support for Solana.

Some AI assistance mainly around tests/docstrings, though all business logic is handrolled.

This comes with a full refactor of the EVM API code, as there was a lot of RPC behaviour that was generically reusable by Solana (and hopefully by other chains too).

# Notes:
* I tried to replicate the capabilities available on the EVM side, which means Solana is subject to the same limitations. For transaction sending, transactions are expected to be fully constructed and signed outside of the rule. 
* I carried over a limitation/potential issue from EVM as it's not entirely clear to me whether it's deliberate. Only non-2XX responses from the RPC are treated as failure, to be consistent with EVM, where 200 responses with a JSON-RPC error body are bubbled up as a success. This is not ideal in Solana as there are some retryable errors that are returned in this way, i.e. -32005 (node is unhealthy/behind) or. -32004 (block not available). LMK what you think about the alternative of turning these cases into errors, and whether that should apply to EVM too.
* I added some unit tests using `surfpool` as a lightweight test validator. This means we add quite a few `dev-dependencies`, which blow up `Cargo.lock`, but naturally these should have no effect on compilation times outside the tests. Still, let me know if it's overkill.
* I'm not sure about the best approach to test this e2e. Within this repo boundary, I'm not seeing much testing done for the EVM contract reader/writer, so I'm keeping things similar.

# On solana specifics

Naturally not all concepts from EVM translate 1:1 to Solana. All behaviour that's available through the EVM guest wrappers should be available on Solana, though the translation may not always be obvious. There are also solana-specific capabilities such as retrieving accounts owned by a program (with filters) that are quite powerful but require careful filtering not to blow up the response size.

Note also how commitment level (i.e. `committed`, `finalized`) is a global config flag, not a granular parameter per method. I think this is the simplest approach and easiest way to prevent mistakes, and I can't easily think of a rule that would want to mix and match.

### DIFF
--- a/runtime/plaid-stl/src/blockchain/solana/mod.rs
+++ b/runtime/plaid-stl/src/blockchain/solana/mod.rs
@@ -251,7 +251,7 @@ pub fn get_multiple_accounts(
 /// unfiltered scan, which can return a large payload — see [`RETURN_BUFFER_SIZE`].
 pub fn get_program_accounts(
     program_id: &UnvalidatedPubkey,
-    filters: ProgramAccountsFilters,
+    filters: &ProgramAccountsFilters,
     cluster: Cluster,
 ) -> Result<SolanaRpcResponse, PlaidFunctionError> {
     extern "C" {
@@ -260,7 +260,7 @@ pub fn get_program_accounts(
     let request = serialize_request(&GetProgramAccountsRequest {
         cluster,
         program_id: Cow::Borrowed(program_id),
-        filters,
+        filters: Cow::Borrowed(filters),
     })?;
     let mut buffer = vec![0; RETURN_BUFFER_SIZE];
     let res = unsafe {
--- a/runtime/plaid-stl/src/blockchain/solana/mod.rs
+++ b/runtime/plaid-stl/src/blockchain/solana/mod.rs
@@ -1,6 +1,8 @@
 pub mod types;
 mod utils;
 
+use std::borrow::Cow;
+
 pub use utils::{parse_rpc_response, SolanaError};
 
 use crate::blockchain::solana::types::{
@@ -8,8 +10,8 @@ use crate::blockchain::solana::types::{
     GetMinimumBalanceForRentExemptionRequest, GetMultipleAccountsRequest,
     GetProgramAccountsRequest, GetRecentPrioritizationFeesRequest, GetSignatureStatusesRequest,
     GetSignaturesForAddressRequest, GetTokenAccountsByOwnerRequest, GetTransactionRequest,
-    ProgramAccountsFilters, Pubkey, PubkeyRequest, SendTransactionRequest, Signature,
-    SolanaRpcResponse,
+    ProgramAccountsFilters, PubkeyRequest, SendTransactionRequest, SolanaRpcResponse,
+    UnvalidatedPubkey, UnvalidatedSignature,
 };
 use crate::PlaidFunctionError;
 use serde::Serialize;
@@ -22,7 +24,7 @@ const RETURN_BUFFER_SIZE: usize = 1024 * 1024; // 1 MiB
 
 /// Submits a fully-signed, base64-encoded transaction (`sendTransaction`).
 pub fn send_signed_transaction(
-    transaction: String,
+    transaction: impl AsRef<str>,
     cluster: Cluster,
 ) -> Result<SolanaRpcResponse, PlaidFunctionError> {
     extern "C" {
@@ -30,7 +32,7 @@ pub fn send_signed_transaction(
     }
     let request = serialize_request(&SendTransactionRequest {
         cluster,
-        transaction,
+        transaction: Cow::Borrowed(transaction.as_ref()),
     })?;
     let mut buffer = vec![0; RETURN_BUFFER_SIZE];
     let res = unsafe {
@@ -46,13 +48,16 @@ pub fn send_signed_transaction(
 
 /// Returns the lamport balance of an account (`getBalance`).
 pub fn get_balance(
-    pubkey: Pubkey,
+    pubkey: &UnvalidatedPubkey,
     cluster: Cluster,
 ) -> Result<SolanaRpcResponse, PlaidFunctionError> {
     extern "C" {
         new_host_function_with_error_buffer!(blockchain_solana, get_balance);
     }
-    let request = serialize_request(&PubkeyRequest { cluster, pubkey })?;
+    let request = serialize_request(&PubkeyRequest {
+        cluster,
+        pubkey: Cow::Borrowed(pubkey),
+    })?;
     let mut buffer = vec![0; RETURN_BUFFER_SIZE];
     let res = unsafe {
         blockchain_solana_get_balance(
@@ -67,13 +72,16 @@ pub fn get_balance(
 
 /// Returns all information associated with an account (`getAccountInfo`).
 pub fn get_account_info(
-    pubkey: Pubkey,
+    pubkey: &UnvalidatedPubkey,
     cluster: Cluster,
 ) -> Result<SolanaRpcResponse, PlaidFunctionError> {
     extern "C" {
         new_host_function_with_error_buffer!(blockchain_solana, get_account_info);
     }
-    let request = serialize_request(&PubkeyRequest { cluster, pubkey })?;
+    let request = serialize_request(&PubkeyRequest {
+        cluster,
+        pubkey: Cow::Borrowed(pubkey),
+    })?;
     let mut buffer = vec![0; RETURN_BUFFER_SIZE];
     let res = unsafe {
         blockchain_solana_get_account_info(
@@ -142,13 +150,16 @@ pub fn get_transaction_count(cluster: Cluster) -> Result<SolanaRpcResponse, Plai
 
 /// Returns details for a confirmed transaction by signature (`getTransaction`).
 pub fn get_transaction(
-    signature: Signature,
+    signature: &UnvalidatedSignature,
     cluster: Cluster,
 ) -> Result<SolanaRpcResponse, PlaidFunctionError> {
     extern "C" {
         new_host_function_with_error_buffer!(blockchain_solana, get_transaction);
     }
-    let request = serialize_request(&GetTransactionRequest { cluster, signature })?;
+    let request = serialize_request(&GetTransactionRequest {
+        cluster,
+        signature: Cow::Borrowed(signature),
+    })?;
     let mut buffer = vec![0; RETURN_BUFFER_SIZE];
     let res = unsafe {
         blockchain_solana_get_transaction(
@@ -167,7 +178,7 @@ pub fn get_transaction(
 /// that have left the recent status cache; this is expensive, so leave it
 /// `false` for the common "is this recent signature confirmed?" check.
 pub fn get_signature_statuses(
-    signatures: Vec<Signature>,
+    signatures: &[UnvalidatedSignature],
     search_transaction_history: bool,
     cluster: Cluster,
 ) -> Result<SolanaRpcResponse, PlaidFunctionError> {
@@ -176,7 +187,7 @@ pub fn get_signature_statuses(
     }
     let request = serialize_request(&GetSignatureStatusesRequest {
         cluster,
-        signatures,
+        signatures: Cow::Borrowed(signatures),
         search_transaction_history,
     })?;
     let mut buffer = vec![0; RETURN_BUFFER_SIZE];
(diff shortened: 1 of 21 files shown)


obelisk:
This adds a huge dependency tree; could we add a feature so deployments that don't need these APIs don't need the deps?

PabloMansanet:
> This adds a huge dependency tree; could we add a feature so deployments that don't need these APIs don't need the deps?

Sure, they're just for the tests anyway so no problem gating them behind a feature (or removing the tests altogether if these kind of unit tests are not how things are done around here; they've served their purpose).

EDIT: Unfortunately a feature flag won't affect the deps, so to reduce that `Cargo.lock` a bit we'd need to get rid of the tests. I'll comment them out for now and document why, and we can remove them fully later.

obelisk:
Ok makes sense. @wbssbw can you review this and run the tests yourself as well

runtime/plaid-stl/src/blockchain/solana/mod.rs
```diff
@@ -0,0 +1,475 @@
...
+            buffer.as_mut_ptr(),
+            RETURN_BUFFER_SIZE,
+        )
+    };
+    finish(res, buffer)
+}
+
+/// Returns the lamport balance of an account (`getBalance`).
+pub fn get_balance(
+    pubkey: Pubkey,
```
obelisk: Do you want to consume the object instead of taking a reference? Seems like having the object be clone (or having `PubkeyRequest` take references) might be more ergonomic?

runtime/plaid-stl/src/blockchain/solana/mod.rs
```diff
@@ -0,0 +1,475 @@
...
+    finish(res, buffer)
+}
+
+/// Returns the processing statuses of a batch of signatures (`getSignatureStatuses`).
+///
+/// Set `search_transaction_history` to `true` to scan the full ledger for signatures
+/// that have left the recent status cache; this is expensive, so leave it
+/// `false` for the common "is this recent signature confirmed?" check.
+pub fn get_signature_statuses(
+    signatures: Vec<Signature>,
```
obelisk: I think this should take `&[Signature]` which is more generic than `Vec`. Is there a reason you want to take in a Vec and consume it?

runtime/plaid-stl/src/blockchain/solana/mod.rs
```diff
@@ -0,0 +1,475 @@
...
+            buffer.as_mut_ptr(),
+            RETURN_BUFFER_SIZE,
+        )
+    };
+    finish(res, buffer)
+}
+
+/// Simulates a signed transaction without submitting it (`simulateTransaction`).
+pub fn simulate_transaction(
+    transaction: String,
```
obelisk: Why a `String` when you could take an `&str`?

runtime/plaid-stl/src/blockchain/solana/types.rs
```diff
@@ -0,0 +1,328 @@
+use serde::{Deserialize, Serialize};
+use std::fmt::{self, Display};
+use std::str::FromStr;
+
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
```
obelisk: Why? the following are all capitalized with no capital letters within? Won't this do nothing? If not, what does it do?

runtime/plaid-stl/src/blockchain/solana/types.rs
```diff
@@ -0,0 +1,328 @@
...
+}
+
+/// A base58-encoded Solana account address (a 32-byte public key). Unvalidated
+/// (the newtype is just to disambiguate with other strings: it's the caller's
+/// responsibility to encode as 32 byte b58).
+#[derive(Clone, Debug, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct Pubkey(String);
+
+impl Pubkey {
```
obelisk: Perhaps this is my security engineer side coming out but I think this should be called `UnverifiedPubkey` or the `new` function should return `Result` if what's passed is invalid. Otherwise, we just leave open opportunity for people to do it wrong and duplicate checking code across modules.

runtime/plaid-stl/src/blockchain/solana/types.rs
```diff
@@ -0,0 +1,328 @@
...
+impl From<Pubkey> for String {
+    fn from(p: Pubkey) -> String {
+        p.0
+    }
+}
+
+/// A base58-encoded Solana transaction signature (64 bytes).
+///
+/// The signature analog of [`Pubkey`]: a disambiguation-only newtype that performs
+/// **no validation**. See [`Pubkey`] for the rationale.
```
obelisk, line 86: Again, this feels wrong. `UnverifiedSignature` perhaps? Maybe more information can be provided on why you want this types named this way?

runtime/plaid/src/apis/blockchain/common/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+    HashMap::<String, ChainConfig>::deserialize(deserializer)?
+        .into_iter()
+        .map(|(key, config)| Ok((key.parse().map_err(de::Error::custom)?, config)))
+        .collect()
+}
+
+#[derive(Deserialize)]
+pub struct ChainConfig {
+    /// The list of nodes for this chain
+    pub(crate) nodes: Vec<NodeConfig>,
```
obelisk: Lots of `pub(crate)`, why this instead of pub for these structures?

runtime/plaid/src/apis/blockchain/common/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+        selector: &NodeSelector,
+        identifier: C::Identifier,
+        json_rpc_request: JsonRpcRequest<'_, M>,
+        module: Arc<PlaidModule>,
+    ) -> Result<String, ApiError> {
+        let mut last_error = BlockchainError::AllNodesFailed;
+
+        debug!(
+            "Module [{module}] is attempting to call [{}] on chain [{}]",
+            json_rpc_request.method, identifier,
```
obelisk, line 209: Nit (because we need to validate the whole code base for this): If this data is passed right through from the rule, it's possible this might be log injection.

obelisk (changes requested):
Going to let @wbssbw review this as well but from my perspective I think the `stl` code should take more references rather than consuming objects from my current understanding.

If there's a good reason the way it is please correct me. Right now I think we'll end up calling clone in rules a lot when we could avoid it.

runtime/plaid-stl/src/blockchain/solana/types.rs
PabloMansanet: Ah, my bad, this was from when I used to call `Mainnet` `MainnetBeta`, which is more accurate to the Solana ecosystem but confusing in its own way. I'll remove it.

runtime/plaid/src/apis/blockchain/common/mod.rs
PabloMansanet: In general as I was refactoring the EVM module into `common`, I wanted to expose some fields only to the `evm` and `solana` api modules while still keeping `common` public, but in hindsight it's a bit unnecessary as `plaid` isn't consumed as a library anyway, so I'll keep it simple and make it all pub.

runtime/plaid/src/apis/blockchain/common/mod.rs
PabloMansanet: This snippet is preexisting (as is) from the previous EVM `execute_rpc_call` method. The only difference form before is that `method` is now any implementer of Serialize rather than the specific EVM enum, but in practice we only pass the two RpcRequest enum from the respective chain families.

Not sure what needs changing, just removing the log altogether?

runtime/plaid-stl/src/blockchain/solana/mod.rs
```diff
@@ -0,0 +1,498 @@
...
+            RETURN_BUFFER_SIZE,
+        )
+    };
+    finish(res, buffer)
+}
+
+/// Returns the lamport balance of an account (`getBalance`).
+pub fn get_balance(
+    pubkey: &UnvalidatedPubkey,
+    cluster: Cluster,
```
obelisk, line 52: Reason to consume `Cluster`?

runtime/plaid-stl/src/blockchain/solana/types.rs
```diff
@@ -0,0 +1,328 @@
...
+///
+/// Mirrors the EVM `SendRawTransactionRequest`: the caller is responsible for
+/// building, signing, and encoding the transaction; the host simply relays it
+/// to the node's `sendTransaction` RPC method.
+#[derive(Serialize, Deserialize)]
+pub struct SendTransactionRequest<'a> {
+    /// Target cluster.
+    pub cluster: Cluster,
+    /// Base64-encoded, fully-signed transaction.
+    pub transaction: Cow<'a, str>,
```
obelisk, line 176: I don't have a problem with Cow, but why it instead of `&'a str`? At least lets document in the code why

runtime/plaid-stl/src/blockchain/solana/types.rs
```diff
@@ -0,0 +1,328 @@
...
+
+/// Submits a fully-signed transaction to a cluster.
+///
+/// Mirrors the EVM `SendRawTransactionRequest`: the caller is responsible for
+/// building, signing, and encoding the transaction; the host simply relays it
+/// to the node's `sendTransaction` RPC method.
+#[derive(Serialize, Deserialize)]
+pub struct SendTransactionRequest<'a> {
+    /// Target cluster.
+    pub cluster: Cluster,
```
obelisk, line 162: Then this could also be an `&'a Cluster` or `&'b Cluster`?

runtime/plaid-stl/src/blockchain/solana/mod.rs
```diff
@@ -0,0 +1,498 @@
...
+
+/// Enumerates the accounts owned by a program (`getProgramAccounts`).
+///
+/// Pass [`ProgramAccountsFilters`] to scope the scan by account size / `memcmp` and to
+/// project a slice of each account's data. An empty (default) filter set performs an
+/// unfiltered scan, which can return a large payload — see [`RETURN_BUFFER_SIZE`].
+pub fn get_program_accounts(
+    program_id: &UnvalidatedPubkey,
+    filters: ProgramAccountsFilters,
+    cluster: Cluster,
```
obelisk: More types that consume? Any particular reason?

runtime/plaid/src/apis/blockchain/common/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    parse_duration,
+};
+use http::{header::CONTENT_TYPE, HeaderMap, HeaderValue, StatusCode};
+use reqwest::Client;
+use serde::{de, Deserialize, Serialize};
+use std::{
+    collections::HashMap, fmt::Display, hash::Hash, str::FromStr, sync::Arc, time::Duration,
+};
+
+pub trait ChainFamily {
```
obelisk, line 35: Documentation on the trait please :)

obelisk (changes requested):
The only real blocking thing from me is more documentation on why we're using `Cow`. I think the intention is so the STL has the option of making some changes to ease the ergonomics of the API but I'd like to see some more documentation on that.

runtime/plaid-stl/src/blockchain/solana/mod.rs
PabloMansanet: `Cluster` is a raw `enum` with three variants, and it's `Copy`. That's just 1 byte of memory; more efficient than passing `&Cluster` :)

runtime/plaid-stl/src/blockchain/solana/types.rs
PabloMansanet: Will add some documents at the top of the file to justify!

runtime/plaid-stl/src/blockchain/solana/types.rs
PabloMansanet: Ditto on the Cluster, it's a one byte copy type, no need to hold a reference to it.

runtime/plaid-stl/src/blockchain/solana/mod.rs
PabloMansanet: Cluster is as above; `filters` is an oversight: I'm Cow-ing it in the next commit.

wbssbw (approved):
Looks good to me! Nice work. Let's wait for @obelisk to take a final pass and then we're good to merge

obelisk (approved):

PR #325, opened 2026-06-19
Add GH API to create deploy keys

michelemin:
Also reorganize STL code into a dedicated module.

### DIFF
--- a/runtime/plaid-stl/src/github/deploy_keys.rs
+++ b/runtime/plaid-stl/src/github/deploy_keys.rs
@@ -1,0 +1,94 @@
+use std::fmt::Display;
+
+use crate::{
+    github::{CreateDeployKeyParams, DeleteDeployKeyParams, GithubApiWrapper},
+    PlaidFunctionError,
+};
+
+/// Delete a deploy key with given ID from a given repository.
+/// ## Arguments
+/// * `client_id` - Selects which configured GitHub client to use (supports multiple clients).
+/// * `owner` - The account owner of the repository.
+/// * `repo` - The name of the repository.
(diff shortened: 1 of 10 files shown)


runtime/plaid-stl/src/github/deploy_keys.rs
```diff
@@ -0,0 +1,94 @@
...
+    // There was an error with the Plaid system. Maybe the API is not
+    // configured.
+    if res < 0 {
+        return Err(res.into());
+    }
+
+    Ok(())
+}
+
+/// Create a deploy key with given ID for a given repository.
```
obelisk: "Using the given `client_id`" probably more clear

runtime/plaid/src/apis/github/deploy_keys.rs
```diff
@@ -44,4 +45,61 @@ impl Github {
...
+            read_only: bool,
+        }
+
+        let body = Body {
+            title: title.to_string(),
+            key: key.to_string(),
+            read_only,
+        };
+
+        info!("Creating deploy key with title [{title}] for repo {owner}/{repo}. {final_part}");
```
obelisk: I would probably have this be:
```
        let perm = match read_only {
            true => "-only",
            false => "/write",
        };
```
`info!("Creating read{perm} deploy key...")`

obelisk (dismissed):
Couple nits but not enough to block

runtime/plaid/src/apis/github/deploy_keys.rs
michelemin: Very elegant!

timweri (approved):

PR #326, opened 2026-06-24
rate-limit-safe outbound message queue for slack

nazriz:
Bursts of Slack messages to one channel hit Slack's ~1/sec/channel limit and get dropped. This adds a paced, durable outbound path:

- Non-blocking `slack_enqueue_message` host fn backed by a per-channel queue
- Background task delivers at <=1/sec/channel, FIFO, honouring Retry-After
- Persists across restarts; dead-letters after capped retries
- Posted ref written back to caller's storage for later edits
- Opt-in; `post_message` unchanged

### DIFF
--- a/runtime/plaid-stl/src/slack/mod.rs
+++ b/runtime/plaid-stl/src/slack/mod.rs
@@ -278,6 +278,63 @@ pub fn post_message_with_blocks_and_attachments_unfurl(
     serde_json::from_str(&res).map_err(|_| PlaidFunctionError::Unknown)
 }
 
+/// Enqueue a Block Kit message for paced, rate-limit-safe delivery by the runtime's
+/// outbound Slack queue, instead of posting inline. Non-blocking: returns as soon as
+/// the runtime has durably queued it. The runtime posts at ≤1 message/sec/channel
+/// (Slack's `chat.postMessage` limit), in order, retrying on rate limits, and writes
+/// the resulting `{ok, channel, ts}` ref into this rule's private storage under
+/// `ref_key` once posted — read it later (e.g. to update the message) with
+/// `plaid::storage::get(ref_key)`.
+pub fn enqueue_message_with_blocks_and_attachments_unfurl(
+    bot: &str,
+    channel: &str,
+    blocks: &str,
+    attachments: &str,
+    unfurl_links: Option<bool>,
+    unfurl_media: Option<bool>,
+    ref_key: &str,
+) -> Result<(), PlaidFunctionError> {
+    extern "C" {
+        new_host_function_with_error_buffer!(slack, enqueue_message);
+    }
+
+    let message = SlackMessageWithBlocksAndAttachments {
+        channel: channel.to_owned(),
+        blocks: optional_blocks(blocks)?,
+        attachments: serde_json::from_str(attachments)
+            .map_err(|_| PlaidFunctionError::ErrorCouldNotSerialize)?,
+        unfurl_links,
+        unfurl_media,
+    };
+
+    let params = serde_json::json!({
+        "bot": bot,
+        "channel": channel,
+        "body": serde_json::to_string(&message).map_err(|_| PlaidFunctionError::ErrorCouldNotSerialize)?,
+        "ref_key": ref_key,
+    })
+    .to_string();
+
+    // Success returns an empty body; a small buffer is plenty.
+    const RETURN_BUFFER_SIZE: usize = 1024;
+    let mut return_buffer = vec![0; RETURN_BUFFER_SIZE];
+
+    let res = unsafe {
+        slack_enqueue_message(
+            params.as_bytes().as_ptr(),
+            params.as_bytes().len(),
+            return_buffer.as_mut_ptr(),
+            RETURN_BUFFER_SIZE,
+        )
+    };
+
+    if res < 0 {
+        return Err(res.into());
+    }
+
+    Ok(())
+}
+
 /// Slack API response for message operations (chat.postMessage, chat.update).
 #[derive(Serialize, Deserialize)]
 pub struct SlackMessageResponse {
--- a/runtime/plaid/src/apis/slack/api.rs
+++ b/runtime/plaid/src/apis/slack/api.rs
@@ -37,6 +37,18 @@ struct GenericSlackResponse {
     ok: bool,
 }
 
+/// Parameters for `enqueue_message`: a pre-rendered `chat.postMessage` body plus
+/// the channel (for per-channel queueing/pacing) and the storage key under which
+/// the posted message ref should be written back.
+#[derive(serde::Deserialize)]
+struct EnqueueMessage {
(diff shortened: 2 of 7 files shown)


runtime/plaid-stl/src/slack/mod.rs
```diff
@@ -278,6 +278,63 @@ pub fn post_message_with_blocks_and_attachments_unfurl(
...
+
+    let params = serde_json::json!({
+        "bot": bot,
+        "channel": channel,
+        "body": serde_json::to_string(&message).map_err(|_| PlaidFunctionError::ErrorCouldNotSerialize)?,
+        "ref_key": ref_key,
+    })
+    .to_string();
+
+    // Success returns an empty body; a small buffer is plenty.
```
michelemin, line 318: Then let's consider whether this should have a return buffer at all. We can use `new_host_function!` and simplify the code, if we are not getting anything interesting back from the host.

runtime/plaid/src/apis/slack/queue.rs
```diff
@@ -0,0 +1,337 @@
...
+use super::SLACK_POST_MESSAGE_URL;
+
+/// Storage namespace for pending outbound posts. Host-internal (the runtime owns
+/// it directly), so it needs no shared-DB configuration.
+pub const QUEUE_NAMESPACE: &str = "__slack_outbound_queue__";
+/// Storage namespace for posts that exhausted their attempt budget.
+pub const DLQ_NAMESPACE: &str = "__slack_outbound_dlq__";
+/// Failed post attempts before an entry is dead-lettered. Only genuine post
+/// failures count; entries waiting their turn in line do not.
+pub const MAX_ATTEMPTS: u32 = 15;
```
michelemin, line 35: Nit - We could make this configurable, with a default.

runtime/plaid/src/apis/slack/queue.rs
```diff
@@ -0,0 +1,337 @@
...
+    /// Failed post attempts so far.
+    pub attempts: u32,
+    /// Enqueue timestamp (epoch nanos) — orders the per-channel FIFO via the key.
+    pub seq: u128,
+}
+
+/// Storage key: `{channel}:{seq}` zero-padded so lexicographic order is FIFO.
+pub fn storage_key(post: &QueuedPost) -> String {
+    format!("{}:{:039}", post.channel, post.seq)
+}
```
michelemin, line 64: This seems more natural as an `impl` on `QueuedPost`.

runtime/plaid/src/apis/slack/api.rs
```diff
@@ -187,6 +199,48 @@ impl Slack {
         }
     }
 
+    /// Enqueue a `chat.postMessage` for paced, rate-limit-safe delivery by the
+    /// background outbound queue. Non-blocking: it persists the entry and returns
+    /// immediately; the drain task posts it at ≤1/sec/channel and writes the
+    /// resulting `{ok, channel, ts}` ref into the calling rule's private storage
```
obelisk, line 205: This is an interesting approach but it creates a new dependency between API calls and storage set ups. It also will make it difficult to reason about storage limits for people who use those.

runtime/plaid-stl/src/slack/mod.rs
```diff
@@ -278,6 +278,63 @@ pub fn post_message_with_blocks_and_attachments_unfurl(
...
 }
 
+/// Enqueue a Block Kit message for paced, rate-limit-safe delivery by the runtime's
+/// outbound Slack queue, instead of posting inline. Non-blocking: returns as soon as
+/// the runtime has durably queued it. The runtime posts at ≤1 message/sec/channel
+/// (Slack's `chat.postMessage` limit), in order, retrying on rate limits, and writes
+/// the resulting `{ok, channel, ts}` ref into this rule's private storage under
+/// `ref_key` once posted — read it later (e.g. to update the message) with
+/// `plaid::storage::get(ref_key)`.
+pub fn enqueue_message_with_blocks_and_attachments_unfurl(
```
obelisk, line 288: This is setting a precedent where we need to duplicate our whole API to allow non-blocking which is going to significantly increase the maintenance burden.

runtime/plaid/src/apis/slack/api.rs
```diff
@@ -37,6 +37,18 @@ struct GenericSlackResponse {
...
+/// Parameters for `enqueue_message`: a pre-rendered `chat.postMessage` body plus
+/// the channel (for per-channel queueing/pacing) and the storage key under which
+/// the posted message ref should be written back.
+#[derive(serde::Deserialize)]
+struct EnqueueMessage {
+    bot: String,
+    channel: String,
+    /// Fully serialized `chat.postMessage` JSON body.
+    body: String,
+    ref_key: String,
```
obelisk, line 49: This design might be too limited as after this, there will definitely be an ask for this to write to shared DBs so other rules could access it. But I don't think this is the right path forward as it creates some difficult dependencies across the Plaid runtime (API probably shouldn't rely on a database, and the kind of database also changes the properties of this).

runtime/plaid/src/apis/slack/mod.rs
```diff
@@ -1,16 +1,27 @@
...
+
+use crate::storage::Storage;
+
 use super::default_timeout_seconds;
 
+pub(crate) use queue::QueuedPost;
+
+/// Slack `chat.postMessage` endpoint, used by the outbound drain task.
+pub(crate) const SLACK_POST_MESSAGE_URL: &str = "https://slack.com/api/chat.postMessage";
+
```
obelisk, line 24: This is code duplication from the API and we would likely want the whole Slack API (or all APIs to have this new feature)

runtime/plaid/src/apis/slack/queue.rs
```diff
@@ -0,0 +1,337 @@
...
+
+use crate::storage::Storage;
+
+use super::SLACK_POST_MESSAGE_URL;
+
+/// Storage namespace for pending outbound posts. Host-internal (the runtime owns
+/// it directly), so it needs no shared-DB configuration.
+pub const QUEUE_NAMESPACE: &str = "__slack_outbound_queue__";
+/// Storage namespace for posts that exhausted their attempt budget.
+pub const DLQ_NAMESPACE: &str = "__slack_outbound_dlq__";
```
obelisk, line 32: Databases don't have to exist

obelisk (changes requested):
I don't think this is the right way to solve this problem. I've left some comments on specifics but broadly this needs to be solved holistically at the API structure level, not just for a particular API within a sub-module.

This is something we've been discussing for a long time and have some ideas on how to implement it and it seems like now is the time. Functionally, we need to find a way to implement async/await across the boundary. With this method, we're going to end up with a quite unweildly API with a lot of duplication and queues that each need management and draining to ensure proper behaviour.

This also means all existing rules need to be ported to get this new behaviour rather than just getting increased reliability from an STL update.

PR #329, opened 2026-06-30
Support updating files in GH

michelemin:
The [endpoint](https://docs.github.com/en/rest/repos/contents?apiVersion=2026-03-10#create-or-update-file-contents) to create and update a file is the same, but updating also requires passing the SHA of the to-be-updated file.

This PR adds support for that, by adding an `update_file` method that transparently fetches the file, gets its SHA, and passes it to the endpoint via the host.

Along the way, we also make `reference` optional when fetching a file: according to the API docs, not passing a reference means fetching from the default branch, which is what we want. Note - THIS IS A BREAKING CHANGE for modules.

### DIFF
--- a/runtime/plaid-stl/Cargo.toml
+++ b/runtime/plaid-stl/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "plaid_stl"
-version = "0.44.1"
+version = "0.44.2"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/runtime/plaid-stl/Cargo.toml
+++ b/runtime/plaid-stl/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "plaid_stl"
-version = "0.44.2"
+version = "0.45.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/runtime/plaid/src/apis/github/repos.rs
+++ b/runtime/plaid/src/apis/github/repos.rs
@@ -650,7 +650,6 @@ impl Github {
         let path = self.validate_path(&request.params.path)?;
         let file_hash = sha256_hex(&request.params.content);
 
-        info!("Creating file with hash [{file_hash}] at [{path}] in repository [{owner}/{repo}] on behalf of [{module}]");
         let address = format!("/repos/{owner}/{repo}/contents/{path}");
 
         let mut body = json!({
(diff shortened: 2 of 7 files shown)


runtime/plaid/src/apis/github/repos.rs
```diff
@@ -204,20 +204,25 @@ impl Github {
...
+                // To validate that the provided ref is valid, we must check that it is either a
+                // commit hash or branch name using the provided validator functions
+                if self.validate_commit_hash(&r).is_err() && self.validate_branch_name(&r).is_err()
+                {
+                    return Err(ApiError::BadRequest);
+                }
+                (format!(" and reference [{r}]"), format!("?ref={r}"))
+            }
+            None => (String::new(), String::new()),
+        };
```
michelemin: We validate the ref if it's given. Otherwise the GH API will default to the repo's default branch.

runtime/plaid-stl/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "plaid_stl"
-version = "0.44.1"
+version = "0.44.2"
```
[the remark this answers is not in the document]
obelisk: @michelemin This is correct. This needs to be version 45

runtime/plaid/src/apis/github/repos.rs
```diff
@@ -204,20 +204,25 @@ impl Github {
...
-            && self.validate_branch_name(&reference).is_err()
-        {
-            return Err(ApiError::BadRequest);
-        }
+        let (ref_log, ref_q_param) = match reference {
+            Some(r) => {
+                // Reference can be commit hash OR a branch name.
+                // To validate that the provided ref is valid, we must check that it is either a
+                // commit hash or branch name using the provided validator functions
+                if self.validate_commit_hash(&r).is_err() && self.validate_branch_name(&r).is_err()
```
obelisk, line 213: You need them to both be errors?

obelisk (changes requested):

runtime/plaid/src/apis/github/repos.rs
michelemin: Yes, improved comments but we want to fail if both validations fail, i.e., if both are errors

runtime/plaid/src/apis/github/repos.rs
```diff
@@ -202,22 +202,28 @@ impl Github {
...
-        {
-            return Err(ApiError::BadRequest);
-        }
+        // If a reference is given (and passes validation), we will log about it and append it to the request URL as a query parameter.
+        // If it is not given, we will omit that part of the log and append nothing to the URL.
+        let (ref_log, ref_q_param) = match request.params.reference {
+            Some(r) => {
+                // According to https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content, reference can be
+                // a commit SHA, a branch name, or a tag name. We will validate that it is either a valid commit hash or a valid branch name.
+                // This means that we try validating both ways: if both fail, we return an error. If either succeeds, we continue.
```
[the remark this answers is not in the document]
obelisk: This seems sensible

obelisk (approved):

PR #330, opened 2026-07-01
slack: pace chat.postMessage to <=1/sec per channel

nazriz:
During alert bursts the notifier fires many chat.postMessage calls to the same channel
at once. Slack allows ~1/sec/channel, so the excess returns HTTP 429 and those alerts
are dropped.

To solve for this: Pace chat.postMessage per channel inside the Slack API: before posting, wait until the
channel's next 1s slot, then reserve the following one. Same-channel posts end up ~1s
apart; different channels are independent (the pacing lock is released before the wait).


nazriz:
Closed in favour of: https://github.com/obelisk/plaid/pull/332

PR #332, opened 2026-07-02
slack: schedule messages instead of dropping them on rate limit

nazriz:
## Problem

During alert bursts the notifier can exceed Slack’s per-channel chat.postMessage rate limit (~1 message/sec). Slack responds with HTTP 429, and these responses are currently treated as runtime errors, causing alerts to be dropped.

## Solution

Expose Slack’s scheduled message APIs and introduce an optional rate-limit mode where `postMessage` returns a structured ratelimited response instead of failing. 

This allows the PagerDuty Slack notifier to own the retry policy by scheduling messages into future delivery windows, retrying if a scheduling window is full, while keeping the runtime Slack-agnostic. 

Tested with a simulated 300-message alert storm resulted in zero dropped alerts and no runtime error logs.

### DIFF
--- a/runtime/plaid-stl/src/slack/mod.rs
+++ b/runtime/plaid-stl/src/slack/mod.rs
@@ -1117,10 +1117,12 @@ struct ScheduledMessageBody {
 }
 
 /// Schedule a Block Kit message for future delivery (chat.scheduleMessage).
-/// Used to deliver a message that would otherwise be dropped when a channel is
-/// rate limited (see [`SlackDeliveryResponse::rate_limited`]). The runtime just
-/// relays the call — the caller owns the retry/window policy and inspects
-/// [`ScheduleMessageResponse::window_full`] to pick another `post_at`.
+/// The runtime just relays the call — the caller supplies the delivery time and
+/// owns any retry/window policy, inspecting [`ScheduleMessageResponse::window_full`]
+/// to pick another `post_at` if a window is full. One use is delivering a
+/// message that would otherwise be dropped when a channel is rate limited
+/// (see [`SlackDeliveryResponse::rate_limited`]), but scheduling is general —
+/// any caller that wants a message delivered at a specific time can use it.
 /// - `post_at`: Unix timestamp (seconds) to deliver at; must be in the future.
 pub fn schedule_message(
     bot: &str,
--- a/runtime/plaid-stl/src/slack/mod.rs
+++ b/runtime/plaid-stl/src/slack/mod.rs
@@ -928,3 +928,437 @@ pub fn remove_from_channel(bot: &str, channel: &str, user: &str) -> Result<(), P
 
     Ok(())
 }
+
+#[derive(Serialize)]
+struct SlackMessageWithOptions {
+    channel: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    blocks: Option<serde_json::Value>,
+    attachments: serde_json::Value,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    unfurl_links: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    unfurl_media: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    metadata: Option<serde_json::Value>,
+}
+
+/// Optional settings for [`post_message_with_options`].
+#[derive(Default)]
+pub struct PostMessageOptions {
+    /// Whether Slack should unfurl links in the message.
+    pub unfurl_links: Option<bool>,
+    /// Whether Slack should unfurl media in the message.
+    pub unfurl_media: Option<bool>,
+    /// Message metadata JSON: `{"event_type": "...", "event_payload": {...}}`.
+    /// Metadata is invisible to users but is returned by conversations.history,
+    /// letting rules correlate posted messages back to the event that generated
+    /// them (e.g. to recover the `ts` of a message delivered via scheduling).
+    pub metadata: Option<String>,
+}
+
+/// Response from a `post_message` call. On success the message posted
+/// immediately and `ts` is set. If the caller opted in to rate-limit handling
+/// and the channel was rate limited, `ok` is false and `error` is
+/// `"ratelimited"` — the caller should then schedule the message itself via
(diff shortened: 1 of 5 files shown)


obelisk (changes requested):
I think this scheduling should be done at the rule level letting us more quickly respond to changes in the Slack API and keeping the runtime simpler.

nazriz:
I believe the check failed due to network error, job needs to be re-run, but I am unable to do so.

runtime/plaid-stl/src/slack/mod.rs
```diff
@@ -928,3 +928,437 @@ pub fn remove_from_channel(bot: &str, channel: &str, user: &str) -> Result<(), P
...
+    metadata: Option<serde_json::Value>,
+    post_at: u64,
+}
+
+/// Schedule a Block Kit message for future delivery (chat.scheduleMessage).
+/// Used to deliver a message that would otherwise be dropped when a channel is
+/// rate limited (see [`SlackDeliveryResponse::rate_limited`]). The runtime just
+/// relays the call — the caller owns the retry/window policy and inspects
+/// [`ScheduleMessageResponse::window_full`] to pick another `post_at`.
+/// - `post_at`: Unix timestamp (seconds) to deliver at; must be in the future.
```
obelisk, line 1126: Nit. I would say it's more generic than this, there are lots of reasons to schedule a message.

runtime/plaid/src/apis/slack/api.rs
```diff
@@ -162,10 +216,115 @@ impl Slack {
...
+                        ApiError::SlackError(SlackError::UnexpectedPayload(response.clone()))
+                    })?;
+                if !slack_response.ok {
+                    return Err(ApiError::SlackError(SlackError::UnexpectedPayload(
+                        response,
+                    )));
+                }
+                Ok(response)
+            }
+            Ok((429, _)) if schedule_on_ratelimit => Ok(RATE_LIMITED_BODY.to_string()),
```
obelisk: I think this should pull the actual body, otherwise it'll be very confusing to a caller who would likely expect the Slack response to be passed through?

runtime/plaid/src/apis/slack/api.rs
```diff
@@ -37,13 +41,53 @@ struct GenericSlackResponse {
...
+            Self::ScheduleMessage(p) => client
+                .post(format!("{SLACK_API_URL}{api}", api = "chat.scheduleMessage"))
+                .body(p.body.clone())
+                .header("Content-Type", "application/json; charset=utf-8"),
+            Self::DeleteScheduledMessage(p) => client
+                .post(format!(
+                    "{SLACK_API_URL}{api}",
+                    api = "chat.deleteScheduledMessage"
+                ))
+                .body(p.body().unwrap_or_default()) // TODO this is not great: maybe this method should be fallible
```
obelisk, line 74: Yes this should be fallible but I see why you didn't.

obelisk: Let's file an issue for this

runtime/plaid/src/apis/slack/api.rs
```diff
@@ -162,10 +209,119 @@ impl Slack {
...
+    /// rate-limited response body on 429 (which carries `error: "ratelimited"`)
+    /// instead of a hard error, so it can decide how to react — typically by
+    /// scheduling the message with [`Self::schedule_message`]. How to handle the
+    /// rate limit (delay, windows, retries) is intentionally left to the rule,
+    /// not baked into the runtime.
     pub async fn post_message(&self, params: &str, module: Arc<PlaidModule>) -> Result<String> {
         let p: PostMessage = serde_json::from_str(params).map_err(|_| ApiError::BadRequest)?;
+        let bot = p.bot.clone();
+        // Opt-in per call; unset means the historical "429 is an error" behavior.
+        let schedule_on_ratelimit = serde_json::from_str::<ScheduleOptIn>(params)
```
obelisk: I think this variable is a misnomer now.

runtime/plaid/src/apis/slack/api.rs
```diff
@@ -162,10 +209,119 @@ impl Slack {
...
+            Ok((429, response)) if schedule_on_ratelimit => Ok(response),
+            Ok((status, _)) => Err(ApiError::SlackError(SlackError::UnexpectedStatusCode(
+                status,
+            ))),
+            Err(e) => Err(e),
+        }
+    }
+
+    /// Call the Slack chat.scheduleMessage API. The caller supplies a fully
+    /// rendered body (including `post_at`); the runtime just relays it.
```
obelisk, line 269: I'm not a fan of this but since you've hidden the serialization inside the STL we can change it in the future without a breaking change.

The reason I'd opt not to do that is that we may want to introspect what is happening and ensure only approved fields are getting through

obelisk (dismissed):

runtime/plaid/src/apis/slack/api.rs
nazriz: created: https://github.com/obelisk/plaid/issues/339

runtime/plaid/src/apis/slack/api.rs
nazriz: Yeah it is, I've gone ahead and renamed it

runtime/plaid/src/apis/slack/api.rs
```diff
@@ -37,13 +41,53 @@ struct GenericSlackResponse {
...
+                .post(format!("{SLACK_API_URL}{api}", api = "chat.scheduleMessage"))
+                .body(p.body.clone())
+                .header("Content-Type", "application/json; charset=utf-8"),
+            Self::DeleteScheduledMessage(p) => client
+                .post(format!(
+                    "{SLACK_API_URL}{api}",
+                    api = "chat.deleteScheduledMessage"
+                ))
+                .body(p.body().unwrap_or_default()) // TODO this is not great: maybe this method should be fallible
+                .header("Content-Type", "application/json; charset=utf-8"),
```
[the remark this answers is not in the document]
nazriz: https://github.com/obelisk/plaid/issues/339

obelisk (approved):

PR #334, opened 2026-07-02
Improve module load times

wbssbw:
This pull request refactors and optimizes module loading process, focusing on improved concurrency, separation of concerns, and more efficient storage usage lookups. It also includes a version bump and dependency additions.

**Concurrency and Performance Improvements:**

* Module loading now uses parallel iteration (`rayon`) to speed up reading and compiling modules from disk, handling errors and signature verification in parallel.
* Storage namespace byte-size lookups are now performed concurrently (with bounded concurrency using `futures_util`) after modules are loaded, preventing storage backend overload and improving overall load times.

**Code Structure and Responsibility Changes:**

* The logic for compiling modules (`PlaidModule::compile`) is separated from the logic for populating storage usage and logging, clarifying when and how storage information is set. 
* A new method, `log_load_info`, is added to `PlaidModule` for consistent logging of module load details after storage sizes are populated.

### DIFF
--- a/runtime/plaid/src/loader/mod.rs
+++ b/runtime/plaid/src/loader/mod.rs
@@ -590,7 +590,7 @@ pub async fn load(
                 }
             };
 
-
+            info!("Loading module [{filename}]");
             if let Some(signing) = &config.module_signing {
                 if let Err(e) = check_module_signatures(signing, &filename, &module_bytes) {
                     if signing.panic_on_invalid_signature {
@@ -622,6 +622,7 @@ pub async fn load(
 
             let test_mode = config.test_mode && !config.test_mode_exemptions.contains(&filename);
 
+            
             let mut plaid_module = match PlaidModule::compile(
                 &filename,
                 &config.computation_amount,
@@ -653,6 +654,7 @@ pub async fn load(
             plaid_module.secrets = byte_secrets.get(&type_).map(|x| x.clone());
             plaid_module.accessory_data = module_accessory_data(config, &plaid_module.name, &type_);
 
+            info!("Finished loading module [{filename}]");
             Some(plaid_module)
         })
         .collect();
--- a/runtime/plaid/src/loader/mod.rs
+++ b/runtime/plaid/src/loader/mod.rs
@@ -542,7 +542,7 @@ async fn populate_storage_sizes(
             }
         }
     }))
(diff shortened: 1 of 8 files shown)


runtime/plaid/src/loader/mod.rs
```diff
@@ -490,12 +500,62 @@ impl PlaidModules {
...
+const MAX_STORAGE_LOOKUP_CONCURRENCY: usize = 10;
+
+/// Concurrently populate each module's current storage byte usage from the backing store.
+///
+/// Lookups run with bounded concurrency to avoid overwhelming the storage backend. Each byte count
+/// is written straight into its own module. Modules whose lookup fails are dropped (or the whole
+/// load panics) according to `config.panic_on_module_load_failure`.
+async fn populate_storage_sizes(
+    modules: Vec<PlaidModule>,
+    storage: &Arc<Storage>,
```
obelisk: Why take a reference to an `Arc`? just take an `Arc` and the closure can capture it directly.

runtime/plaid/src/loader/mod.rs
```diff
@@ -13,6 +13,10 @@ use std::fmt::{Display, Formatter};
 use std::fs::{self};
 use std::net::SocketAddr;
 use std::sync::{Arc, RwLock};
+
+use futures_util::stream::{self, StreamExt};
+use rayon::prelude::*;
```
obelisk: I'm generally in favour of only including what we need but don't feel that strongly as long as it's not increasing binary size or increasing compile time

runtime/plaid/src/loader/mod.rs
```diff
@@ -510,114 +570,116 @@ pub async fn load(
...
-            Ok(pm) => pm,
-            Err(e) => {
-                if config.panic_on_module_load_failure {
-                    panic!("Module [{filename}] failed to load: {e}")
-                } else {
-                    error!("Module [{filename}] failed to load: {e}. Skipping module load");
-                    continue;
+            let test_mode = config.test_mode && !config.test_mode_exemptions.contains(&filename);
+
+
```
obelisk: No need for a double break

obelisk (changes requested):
Very minor nit but otherwise seems good to me

obelisk (approved):

PR #340, opened 2026-07-08
Record module failures

wbssbw:
This pull request adds new metrics to track module failures in the executor, improving observability for error monitoring. The main change is the introduction of a new `IntCounterVec` metric to count failures per module, along with the necessary code to increment this metric when a module fails during execution.

**Metrics enhancements:**

* Added a new `IntCounterVec` named `module_failures` to the `ModuleExecutionMetrics` struct to count the number of failures per module.
* Registered the new `module_failures` metric with the metrics handle during initialization.
* Implemented a new method `record_module_failure` in `ModuleExecutionMetrics` to increment the failure counter for a specific module.
* Updated the executor logic in `process_message_with_module` to record a module failure using the new metric whenever a module returns a non-zero value (indicating an error).

### DIFF
--- a/runtime/plaid/src/executor/metrics.rs
+++ b/runtime/plaid/src/executor/metrics.rs
@@ -38,8 +38,14 @@ impl ModuleExecutionMetrics {
         )
         .expect("valid metric definition");
 
-        let module_failures = IntCounterVec::new(Opts::new("name", "help"), &["module"])
-            .expect("valid metric definition");
+        let module_failures = IntCounterVec::new(
+            Opts::new(
+                "plaid_module_failures",
+                "Total number of failures per module",
+            ),
+            &["module"],
+        )
+        .expect("valid metric definition");
 
         handle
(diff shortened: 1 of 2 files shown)


runtime/plaid/src/executor/metrics.rs
```diff
@@ -37,16 +38,29 @@ impl ModuleExecutionMetrics {
...
+
         handle
             .register(Box::new(computation_percentage.clone()))
             .expect("expected unique collector");
         handle
             .register(Box::new(execution_duration_seconds.clone()))
             .expect("expected unique collector");
+        handle
+            .register(Box::new(module_failures.clone()))
+            .expect("expected unique collector");
```
obelisk, line 58: This probably should have been caught in a previous review but `expect` could cause a panic. Are we not worried about that?

obelisk (changes requested):

obelisk (approved):

PR #341, opened 2026-07-12
Allow Google Docs modules to write into folders they create

harry-anderson:
## Summary

- Static `rw` ACLs only cover preconfigured Drive folder IDs
- `create_folder` then `copy_file`/`upload_file` into the new folder failed permission checks
- Track creator ownership at runtime and grant the creating module write/read on those folders

This unblocks rules that clone a template workspace by creating a per-report folder and copying template docs into it.

## Test plan

- [x] `cargo test -p plaid --lib apis::gcp::google_docs::tests::permission_checks`
- [ ] Integration: create_folder under configured parent, then copy_file into the new folder ID

### DIFF
--- a/runtime/plaid/src/apis/gcp/google_docs.rs
+++ b/runtime/plaid/src/apis/gcp/google_docs.rs
@@ -1,4 +1,4 @@
-use std::collections::{HashMap, HashSet};
+use std::collections::{HashMap, HashSet, VecDeque};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
@@ -8,7 +8,7 @@ use plaid_stl::gcp::google_docs::{
     UploadFileOutput,
 };
 use pulldown_cmark::{html, Options, Parser};
-use reqwest::{multipart, Client};
+use reqwest::{multipart, Client, Url};
 use serde::Deserialize;
 use serde_json::{json, Value};
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


harry-anderson:
Unblocks bbqd Drive workspace clone:
- bbqd rule: https://github.com/smartcontractkit/plaid-rules/pull/412
- prodsec config: https://github.com/smartcontractkit/plaid-configs/pull/139

runtime/plaid/src/apis/gcp/google_docs.rs
```diff
@@ -102,6 +120,11 @@ impl GoogleDocs {
                     }
                 }
 
+                // Creator of a folder may also read it
+                if self.module_created_folder(&module, resource_id).await {
+                    return Ok(());
+                }
+
```
michelemin: It seems this can be pulled before the `match` as an early return. Unless a conscious decision was made not to lock the mutex before all other avenues have been tried. In which case, I agree the code should remain as is.

runtime/plaid/src/apis/gcp/google_docs.rs
```diff
@@ -40,6 +40,14 @@ pub struct GoogleDocs {
...
     /// Cached Google OAuth Access Token (Token, Expiry)
     access_token: Mutex<Option<(String, Instant)>>,
+    /// Folders created at runtime by modules: `folder_id -> module_name`.
+    ///
+    /// Static config can only grant write access to known folder IDs. When a
+    /// rule creates a new subfolder (e.g. cloning a template workspace), it
+    /// must also be allowed to write into that newly created folder. Tracking
+    /// creator ownership here closes that gap without opening write access to
+    /// arbitrary Drive IDs the puppet account can see.
+    created_folders: Mutex<HashMap<String, String>>,
```
michelemin: I see the point but this does not survive reboots, correct? So there might be a strange situation where something works and then stops, in a way which is unpredictable from the point of view of a rule.

michelemin (changes requested):
Main comment is about reboots.

michelemin (approved):

PR #343, opened 2026-07-17
github: add GitHub API function to submit pull request reviews

nvsriram:
## Summary

This PR introduces a new GitHub API function that enables Plaid modules to submit a review on a pull request.

## Changes

- **STL**: Added `PullRequestReviewEvent` enum, `SubmitPullRequestReviewRequest` struct, `submit_pull_request_review`, and `approve_pull_request` wrapper in `runtime/plaid-stl/src/github/pull_request.rs`.
- **Runtime**: Implemented `submit_pull_request_review` method in the GitHub API handler in `runtime/plaid/src/apis/github/pull_requests.rs`
- **Validation**: Added `validate_review_body` in `runtime/plaid/src/apis/github/validators.rs` to ensure non-empty `body` is provided when `event` is `RequestChanges` or `Comment`
- **Function Registration**: Added function mapping and implementation in `runtime/plaid/src/functions/api.rs`

## Rationale

This functionality is needed to enable automated approval of pull requests from Plaid modules. The implementation uses the GitHub Pull Request Reviews API endpoint `/repos/{owner}/{repo}/pulls/{number}/reviews` with caller-specified `event` and `body`.

### DIFF
--- a/runtime/plaid-stl/src/github/pull_request.rs
+++ b/runtime/plaid-stl/src/github/pull_request.rs
@@ -169,7 +169,7 @@ pub struct ApprovePullRequestRequest {
     /// The name of the repository without the `.git` extension. The name is not case sensitive.
     pub repo: String,
     /// The number of the pull request to approve.
-    pub pull_number: u64,
+    pub number: u32,
     /// An optional comment to leave alongside the approval.
     pub body: Option<String>,
 }
@@ -183,7 +183,7 @@ pub struct ApprovePullRequestRequest {
 /// * `client_id` - Selects which configured GitHub client to use (supports multiple clients).
 /// * `owner` - The account or organization that owns the repository.
 /// * `repo` - The name of the repository.
-/// * `pull_number` - The number of the pull request to approve.
+/// * `number` - The number of the pull request to approve.
 /// * `body` - An optional comment to leave alongside the approval.
 ///
 /// ## Returns
@@ -193,7 +193,7 @@ pub fn approve_pull_request(
     client_id: impl Display,
     owner: impl Display,
     repo: impl Display,
-    pull_number: u64,
+    number: u32,
     body: Option<impl Display>,
 ) -> Result<String, PlaidFunctionError> {
     extern "C" {
@@ -204,7 +204,7 @@ pub fn approve_pull_request(
     let request = ApprovePullRequestRequest {
         owner: owner.to_string(),
         repo: repo.to_string(),
-        pull_number,
+        number,
         body: body.map(|b| b.to_string()),
     };
 
--- a/runtime/plaid-stl/src/github/pull_request.rs
+++ b/runtime/plaid-stl/src/github/pull_request.rs
@@ -187,7 +187,7 @@ pub struct ApprovePullRequestRequest {
 /// * `body` - An optional comment to leave alongside the approval.
 ///
 /// ## Returns
-/// * `Ok(String)` with the raw JSON of the created review, or
+/// * `Ok(())` if the review was successfully submitted, or
 /// * `Err(PlaidFunctionError)` if the request fails.
 pub fn approve_pull_request(
     client_id: impl Display,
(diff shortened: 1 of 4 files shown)


runtime/plaid-stl/src/github/pull_request.rs
```diff
@@ -161,6 +161,81 @@ pub fn pull_request_request_reviewers(
...
+/// * `Err(PlaidFunctionError)` if the request fails.
+pub fn approve_pull_request(
+    client_id: impl Display,
+    owner: impl Display,
+    repo: impl Display,
+    number: u32,
+    body: Option<impl Display>,
+) -> Result<String, PlaidFunctionError> {
+    extern "C" {
+        new_host_function_with_error_buffer!(github, approve_pull_request);
```
michelemin: I wonder if we really care about the response provided by the API and if we should propagate it back to the caller. If not, this can be a simpler `new_host_function!` and we don't need a return buffer at all.

runtime/plaid/src/apis/github/pull_requests.rs
```diff
@@ -73,6 +73,46 @@ impl Github {
...
+        let owner = self.validate_org(&request.params.owner)?;
+        let repo = self.validate_repository_name(&request.params.repo)?;
+        let number = request.params.number;
+
+        info!("Approving pull request [{owner}/{repo}/{number}] on behalf of {module}");
+
+        let mut request_body = json!({ "event": "APPROVE" });
+        if let Some(body) = request.params.body {
+            request_body["body"] = json!(body);
+        }
```
michelemin: This would be cleaner with a dedicated struct (can be declared inside this function too) that has a body which is an `Option<String>` and is not serialized if `None`.

runtime/plaid/src/apis/github/pull_requests.rs
```diff
@@ -73,6 +73,46 @@ impl Github {
         }
     }
 
+    /// Approves a pull request by submitting an `APPROVE` review on it.
+    pub async fn approve_pull_request(
```
michelemin: I think that here, for a small price, we have a chance to support more than just approving. The REST API allows for approving, requesting changes or submitting a comment. So we could provide all these options for little to no extra cost. The only complication is that a body is mandatory when requesting changes or providing a comment. But it seems simple enough.

Then at the STL level we can still expose a simple `approve_pr` which inside calls something like `submit_review(Review::Approve)` or something similar.

runtime/plaid/src/apis/github/pull_requests.rs
```diff
@@ -73,6 +73,46 @@ impl Github {
...
+
+        let owner = self.validate_org(&request.params.owner)?;
+        let repo = self.validate_repository_name(&request.params.repo)?;
+        let number = request.params.number;
+
+        info!("Approving pull request [{owner}/{repo}/{number}] on behalf of {module}");
+
+        let mut request_body = json!({ "event": "APPROVE" });
+        if let Some(body) = request.params.body {
+            request_body["body"] = json!(body);
```
michelemin: Please let's double check because if I am reading the docs correctly, this is a simple String and not a JSON object.

michelemin (changes requested):
Looks good! I would expand the API a bit and double check the POST body. Then we can ship it 🙂

runtime/plaid/src/apis/github/pull_requests.rs
```diff
@@ -73,6 +73,46 @@ impl Github {
...
+        }
+
+        let address = format!("/repos/{owner}/{repo}/pulls/{number}/reviews");
+
+        match self
+            .make_generic_post_request(&request.client_id, address, request_body, module)
+            .await
+        {
+            Ok((status, Ok(body))) => {
+                if status == 200 {
```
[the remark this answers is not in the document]
nvsriram: The endpoint for both API versions ([2022-11-28](https://docs.github.com/en/rest/pulls/reviews?apiVersion=2022-11-28#create-a-review-for-a-pull-request--status-codes) and [2026-03-10](https://docs.github.com/en/rest/pulls/reviews?apiVersion=2026-03-10#create-a-review-for-a-pull-request--status-codes)) only list `200`, `403`, and `422` status codes. Even the example response has status code `200`. There is no mention of status code `201` so this claim seems invalid.

michelemin (approved):
LGTM!

PR #349, opened 2026-08-03
Add runtime APIs to create and revoke a GH installation token

michelemin:
Also reorganize how to make calls to the GH API to support authenticating as an app via a JWT (through a call to Octocrab).

### DIFF
--- a/runtime/plaid-stl/Cargo.toml
+++ b/runtime/plaid-stl/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "plaid_stl"
-version = "46.4.0"
+version = "46.6.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/testing/just_run_please.sh
+++ b/testing/just_run_please.sh
@@ -33,4 +33,4 @@ mkdir -p compiled_modules
 cp modules/target/wasm32-unknown-unknown/release/just_run_please.wasm compiled_modules/just_run_please.wasm
 
 cd runtime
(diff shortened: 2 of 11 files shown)


testing/just_run_please.sh
```diff
@@ -33,4 +33,4 @@ mkdir -p compiled_modules
 cp modules/target/wasm32-unknown-unknown/release/just_run_please.wasm compiled_modules/just_run_please.wasm
 
 cd runtime
-RUST_LOG=plaid=trace cargo run --bin=plaid --no-default-features --features=${COMPILER_BACKEND} -- --config ${CONFIG_WORKING_PATH} --secrets ${SECRETS_WORKING_PATH}
+RUST_LOG=plaid=debug cargo run --bin=plaid --no-default-features --features=${COMPILER_BACKEND} -- --config ${CONFIG_WORKING_PATH} --secrets ${SECRETS_WORKING_PATH}
```
obelisk: Why?

michelemin: Nobody saw anything........ 😇

runtime/plaid-stl/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "plaid_stl"
-version = "46.4.0"
+version = "46.6.0"
```
obelisk, line 3: Is there a reason we're skipping? Not a problem just curious

obelisk (approved):

runtime/plaid-stl/Cargo.toml
michelemin: Yes, because 46.5.0 was cut without bumping TOML files, so now we are jumping to 46.6.0

PR #351, opened 2026-08-18
Feature-gated postgres support

PabloMansanet:
This adds an opinionated, feature-gated approach to supporting Postgres.

I was wondering for some time whether to have the config own the queries, and only allow parametrization, or whether to allow the rules to own the queries. For the kind of situations I'm thinking of using this in (i.e. building intelligent incident response rules that must be pretty flexible in their querying) it wouldn't work to limit the queries to a set curated in the config, so I've made them available to the rule and documented where the risks and limitations are.

I've tested this both on unencrypted and TLS databases (the former demo is in local-dev, but if it feels heavyweight there I can remove it). Works well too with TLS.

I'm not the most familiar with SQL so please shout if I'm doing something silly here, it is very well possible.


runtime/plaid/src/apis/postgres/mod.rs
```diff
@@ -0,0 +1,1129 @@
+use std::{
```
obelisk, line 1: Generally, would break this up into sub-modules where you can break out setting up and servicing API requests.

runtime/plaid/src/apis/postgres/README.md
```diff
@@ -0,0 +1,127 @@
+# Production PostgreSQL API
+
+The PostgreSQL API is a feature-gated, read-only interface for curated SQL in
+Plaid rules. Treat rule-supplied SQL as trusted application code: review it,
```
obelisk, line 4: Nothing in a rule is generally considered trusted, Plaid should function correctly even if rules are trying to harm the runtime.

While saying rules could be malicious is perhaps a bridge too far (it has access to sensitive APIs), we shouldn't trust that the rules do things correctly and need to protect the runtime and associated resources where that makes sense.

In this case, that means not assuming: The SQL is correct, will respect the bounds put on it by the runtime, will follow the correct permissions (read v write), etx.

PR #352, opened 2026-08-19
Add proxy routing mode for blockchain APIs

wbssbw:
## Summary
Introduces a `RoutingMode` enum for the blockchain RPC client that supports two mutually exclusive routing strategies: per-chain node pools (the existing behavior) and a single proxy endpoint. In proxy mode, every chain routes through one URL with the chain identifier appended as a path segment, so arbitrary chain IDs work without per-chain configuration.

## Changes
- **Code**
  - Added `RoutingMode<C, V>` enum in `runtime/plaid/src/apis/blockchain/common/mod.rs` with `Nodes` and `Proxy` variants, deserialized via a `type` tag.
  - Replaced `BlockchainClient.node_selector` and `max_retries` fields with a single `routing` field.
  - Removed `get_node_selector`; `execute_rpc_call` now dispatches on routing mode.
  - Proxy mode performs a single attempt against the proxy URL (retries are expected to be handled by the proxy); node mode retains retry-across-nodes behavior with failure deprioritization.
  - Generalized `deserialize_chains` to be generic over value type and added a `deserialize_url` helper.
  - Updated EVM and Solana call sites to drop the explicit node-selector lookup.
  - Documented the sticky round-robin behavior in `node_selection.rs`.
- **Config**
  - Updated `runtime/plaid/resources/jrp_config/apis.toml` to the new `routing` schema and added commented proxy-mode examples for EVM and Solana.

## Rationale
Proxy mode removes the need to enumerate every chain and its node pool, which is impractical when supporting arbitrary chain identifiers. Keeping node mode as the default preserves existing behavior and avoids forcing a migration on users who manage their own node pools.

## Breaking Changes
- The config schema for blockchain families changes: `chains` and `max_retries` are replaced by a `routing` block with `type = "nodes"` (per-chain `chains` and `max_retries` nested inside) or `type = "proxy"` with a `url`. Existing configs must be updated to the new structure.

### DIFF
--- a/runtime/plaid/src/apis/blockchain/solana/mod.rs
+++ b/runtime/plaid/src/apis/blockchain/solana/mod.rs
@@ -597,7 +597,6 @@ mod tests {
             timeout_millis: Duration::from_secs(10),
             options: SolanaOptions { commitment },
         })
-        .unwrap()
     }
 
     /// Boot an embedded, offline (no mainnet fork) surfnet with a funded payer.
--- a/runtime/plaid/src/apis/blockchain/solana/mod.rs
+++ b/runtime/plaid/src/apis/blockchain/solana/mod.rs
@@ -84,8 +84,6 @@ impl BlockchainClient<Solana> {
             .map_err(BlockchainError::SerdeError)?;
         let cluster = request.cluster;
 
-        let node_selector = self.get_node_selector(cluster)?;
-
         // [base64Tx, { "encoding": "base64", "preflightCommitment": <configured> }]
         let params = (
             request.transaction,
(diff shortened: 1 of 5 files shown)


runtime/plaid/src/apis/blockchain/solana/mod.rs
```diff
@@ -646,11 +590,14 @@ mod tests {
...
-            chains,
+            routing: RoutingMode::Nodes {
+                chains,
+                max_retries: 3,
+            },
             timeout_millis: Duration::from_secs(10),
-            max_retries: 3,
             options: SolanaOptions { commitment },
         })
+        .unwrap()
```
obelisk: We generally try to avoid unwraps, even if it happens at start up, it still makes error handling neater

obelisk (approved):

